PR #20, opened 2024-07-03
address module isolation type errors, add missing peer deps

turbocrime:
build works. haven't tested it


packages/widget/src/hooks/use-assets.ts
```diff
@@ -16,7 +18,7 @@ export function useAssets(options: AssetsRequest = {}) {
         includeCW20Assets: true,
         includeSvmAssets: true,
         ...options,
-      });
+      }) as Promise<Record<string, Asset[]>>;
```
turbocrime, line 21: this works, but is bad. the proper way to do this would be to make sure `skipClient.assets` returns the correct type

packages/widget/src/hooks/use-auto-set-address.ts
```diff
@@ -27,7 +27,7 @@ export const useAutoSetAddress = ({
   signRequired?: boolean;
   chainAddresses: ChainAddresses;
   setChainAddresses: (v: SetChainAddressesParam) => void;
-}) => {
+}): UseQueryResult => {
```
turbocrime, line 30: afaict this query always returns null data

packages/widget/src/hooks/use-balances-by-chain.ts
```diff
@@ -27,7 +27,7 @@ export function useBalancesByChain({
   chain,
   assets,
   enabled = true,
-}: Args) {
+}: Args): UseQueryResult<Record<string, string>> {
```
turbocrime, line 30: it seems like there should be a better type here, but `Record<string, string>` is explicitly used by a `reduce` below.

packages/widget/src/hooks/use-broadcasted-txs.ts
```diff
@@ -13,6 +13,13 @@ interface TransferSequence {
   state: TransferState;
 }
 
+interface TxsStatus {
+  isSuccess: boolean;
+  isSettled: boolean;
+  transferSequence: TransferSequence[];
+  states: StatusState[];
+}
+
```
turbocrime, line 22: declared this interface for convenience, but it could also simply be repeated

packages/widget/src/hooks/use-route.ts
```diff
@@ -44,7 +45,7 @@ export function useRoute({
     splitRoutes: true,
     evmSwaps: true,
   },
-}: UseRouteArgs) {
+}: UseRouteArgs): UseQueryResult<RouteResponse | undefined> {
```
turbocrime, line 48: the query function contains an early void return, so this is correct.

packages/widget/src/lib/wagmi.ts
```diff
@@ -1,5 +1,5 @@
 import { http } from 'viem';
-import { createConfig } from 'wagmi';
+import { Config as WagmiConfig, createConfig } from 'wagmi';
```
turbocrime, line 2: aliased to avoid any confusion with a local type

package.json
```diff
@@ -82,7 +82,7 @@
     "rollup-plugin-peer-deps-external": "^2.2.4",
     "rollup-plugin-postcss": "^4.0.2",
     "rollup-plugin-terser": "^7.0.2",
-    "rollup-plugin-typescript2": "^0.34.0",
+    "rollup-plugin-typescript2": "^0.36.0",
```
turbocrime, line 85: this was the main problem. bugfix now respects exports field of package.json

PR #43, opened 2024-07-09
[API-3093] Support new field 'pretty_name' in /chains response

mattac21:
This PR adds in the `pretty_name` field that we are adding to the api's /v2/info/chains response. 

I updated the `Chain` and `ChainJSON` type, as well as the converters between the two. Let me know if I'm missing places that I need to update.

Also note that the api change is not merged in yet, lets wait to merge until that hits prod.


NotJeremyLiu:
@mattac21 all non-breaking changes are minor version bumps, so you can changeset with that

mattac21:
> @mattac21 all non-breaking changes are minor version bumps, so you can changeset with that

gotcha, added a minor version bump changeset

codingki (approved):

PR #56, opened 2024-07-12
[patch] Fix web components

toddkao:
Remove extra `SwapWidgetProvider` wrapping the web-component
Add new example page injecting vue.js in `example/nextjs` at page `/vue`

### DIFF
--- a/examples/nextjs/pages/vue.tsx
+++ b/examples/nextjs/pages/vue.tsx
@@ -1,0 +1,45 @@
+import { NextPage } from 'next';
+import { useEffect } from 'react';
+import { initializeSwapWidget } from '@skip-go/widget';
+
+const VuePage: NextPage = () => {
+  initializeSwapWidget();
+
+  useEffect(() => {
+    const { createApp } = require('vue/dist/vue.esm-bundler.js');
+    const app = createApp({
+      setup() {
+        const colors = JSON.stringify({
+          primary: '#FF4FFF',
+        });
+        const defaultRoute = JSON.stringify({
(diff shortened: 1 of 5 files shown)


packages/widget/src/ui/WebComponent.tsx
```diff
@@ -32,17 +32,7 @@ const WidgetWithProvider = (props: WebComponentProps) => {
...
-    realProps as WebComponentProps;
-  return (
-    <SwapWidgetProvider
-      toasterProps={toasterProps}
-      endpointOptions={endpointOptions}
-      apiURL={apiURL}
-    >
-      <SwapWidget {...swapWidgetProps} />
-    </SwapWidgetProvider>
-  );
```
toddkao, line 45: This is now a duplicate since `SwapWidget` already comes pre-wrapped with a `Provider`

examples/nextjs/pages/vue.tsx
```diff
@@ -0,0 +1,45 @@
...
+    });
+
+    (app.config.compilerOptions.isCustomElement = (tag: string) =>
+      ['skip-widget'].includes(tag)),
+      app.mount('#vue');
+
+    return () => {
+      app.unmount();
+    };
+  }, []);
```
toddkao, line 40: ideally I would just have a nuxtjs package in `examples/nuxtjs` but I'm running into issues just setting that up because there's issues between `nuxtjs` install and `yarn workspaces`

NotJeremyLiu (approved):

PR #63, opened 2024-07-16
[patch] Add css reset

toddkao:
Add CSS reset, checked that it successfully overrides styles from https://test-skip-widget.vercel.app/

updated:
css reset

old:
bleh

### DIFF
--- a/examples/nextjs/styles/globals.css
+++ b/examples/nextjs/styles/globals.css
@@ -9,5 +9,12 @@ body {
 }
 
 input {
-  background-color: black;
+  background: black;
+}
+
+div,
+p,
+span {
+  border: 2px solid gray;
+  background: gray;
 }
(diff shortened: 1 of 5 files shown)


examples/nextjs/styles/globals.css
```diff
@@ -9,5 +9,13 @@ body {
...
-  background-color: black;
+  background: black;
+}
+
+div,
+p,
+span {
+  color: gray;
+  border: 2px solid gray;
+  background: gray;
```
toddkao, line 20: just used for testing

packages/widget/src/ui/Widget.tsx
```diff
@@ -294,7 +294,10 @@ export const SwapWidgetUI = ({
                 }
               }}
             >
-              <div key={accountStateKey} className="animate-slide-up-and-fade">
+              <div
+                key={accountStateKey}
+                className="animate-slide-up-and-fade text-white"
+              >
```
toddkao, line 300: `div/p/span` are now defaulted to color: black; so this needs to explicitly have text-white (does not inherit it from parent anymore)

PR #69, opened 2024-07-18
Swap widget without providers

toddkao:
Add `SwapWidgetWithoutProviders`

### DIFF
--- a/.changeset/young-seals-repeat.md
+++ b/.changeset/young-seals-repeat.md
@@ -1,5 +1,0 @@
----
-'nextjs': minor
----
-
-Update to add example with explicit provider
--- a/.changeset/young-seals-repeat.md
+++ b/.changeset/young-seals-repeat.md
@@ -1,0 +1,5 @@
+---
+'nextjs': minor
+---
+
+Update to add example with explicit provider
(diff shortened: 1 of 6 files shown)


.changeset/young-seals-repeat.md
```diff
@@ -0,0 +1,5 @@
+---
+'nextjs': minor
+---
+
+Update to add example with explicit provider
```
toddkao, line 5: do we need to update changeset for nextjs project in order for the vercel link to actually update? no right?

examples/nextjs/styles/globals.css
```diff
@@ -16,6 +16,4 @@ div,
 p,
 span {
   color: gray;
-  border: 2px solid gray;
-  background: gray;
```
toddkao, line 20: lol this was messing with the styling on the page, even though it doesn't pierce the shadow dom, given that we wrap things with divs

.changeset/young-seals-repeat.md
codingki: nope we don't need to update this

codingki (approved):

PR #76, opened 2024-07-19
Remove more hard-coded text colors

toddkao:
(I hard-coded this CW-20 badge to show up, it doesn't normally with this token)
akt test

Update color for main button to use theme.textColor
preview route
Update color for max buttons to use theme.textColor
max

Add StyledSwapButton that always uses the textColor as backgroundColor and backgroundColor as textColor (inverse of normal)
swap

Update these two to use borderColor
MORE BORDERS


NotJeremyLiu:
@toddkao okay I'm sorry I made you do this but I actually hate these buttons being black text, for whatever reason it makes it so much harder for my eyes to read them. So, final opinion we keep all text in buttons/pills (these 3 in the screenshot) hardcoded white. If someone wants a white brand color they have to wait for redesign.

Screen Shot 2024-07-18 at 7 35 38 PM

codingki (approved):

codingki (changes requested):
bring back white hardcoded text

codingki (approved):

PR #87, opened 2024-07-24
[FRE 876] Add persistSwapWidgetState prop

toddkao:
Add persistSwapWidgetState prop, which is true by default
Add example in /provider of disabling it

https://github.com/user-attachments/assets/117a8c34-2a3b-438c-b97b-7f012b8f1160


codingki (approved):
nice

PR #91, opened 2024-07-24
[FRE-860] Rip chains codegen 

codingki:


packages/core/src/chains.ts
```diff
@@ -1,32 +1,33 @@
...
+  chain_name: 'dydx',
+  status: 'live',
+  website: 'https://dydx.exchange/',
+  network_type: 'mainnet',
+  pretty_name: 'dYdX Protocol',
+  chain_id: 'dydx-mainnet-1',
+  bech32_prefix: 'dydx',
+  daemon_name: 'dydxprotocold',
+  node_home: '$HOME/.dydxprotocol',
+  key_algos: ['secp256k1'],
```
toddkao, line 17: @codingki you have a linter too that's doing all this i assume lol, it's so annoying

codingki: yeah xd, previously we have a diff eslint config for core and widget. in this monorepo now same

packages/widget/src/hooks/use-chains.ts
```diff
@@ -32,8 +31,8 @@ export function useChains<T = Chain[]>(args: UseChainsQueryArgs<T> = {}) {
         .map((chain): Chain => {
           return {
             ...chain,
-            chainName: chainIdToName[chain.chainID] || chain.chainName,
-            prettyName: chainIdToPrettyName[chain.chainID] || chain.chainName,
+            chainName: chain.chainName || getChain(chain.chainID).chain_name,
+            prettyName: chain.prettyName || getChain(chain.chainID).pretty_name,
```
toddkao, line 35: @codingki I don't know if `chain.chainName` and  `getChain(chain.chainID).chain_name` can both exist
but `getChain(chain.chainID).chain_name || chain.chainName` is what we had before, this looks like you flipped the priority of chain.chainName to be first

codingki: yes intentional. We are trying to not use `getChain`. but will have a different PR for that task

toddkao (approved):

packages/widget/src/hooks/use-chains.ts
toddkao: ok, sounds good

PR #95, opened 2024-07-24
[FRE-884] Add new widget-v2 package

toddkao:
Initialize widget-v2 package with storybook, vite, typescript

Screenshot 2024-07-25 at 5 13 56 PM


codingki:
can you also add dev script in root package.json for widget-v2

toddkao:
> can you also add dev script in root package.json for widget-v2

I just basically ran storybook for the dev script currently, I feel like there's no need to ever really run/test the app outside of storybook since It looks like we can simply mount any react component within storybook

packages/widget-v2/package.json
```diff
@@ -0,0 +1,39 @@
...
+  "name": "widget-v2",
+  "private": true,
+  "version": "0.0.0",
+  "type": "module",
+  "scripts": {
+    "dev": "storybook dev -p 6006",
+    "build": "tsc -b && vite build",
+    "lint": "eslint . --ext ts,tsx --report-unused-disable-directives --max-warnings 0",
+    "preview": "vite preview",
+    "build-storybook": "storybook build"
```
toddkao, line 11: @codingki so basically dev would run the storybook dev

codingki: yes but add it in our root package.json scripts. maybe dev:widgetv2 or something

toddkao: oh i see i misunderstood what you said

codingki (approved):
lgtm

PR #100, opened 2024-07-25
refactor(FRE-880): increase API chains data usage

codingki:
Fee assets, explorerUrl, pretty name now comes from API


packages/widget/src/hooks/use-swap-widget.ts
```diff
@@ -620,11 +618,9 @@ export function useSwapWidget(persistSwapWidgetState = true) {
...
           });
 
-          feeDenomPrices ??= (await getChainFeeAssets(srcChain.chainID)).find(
-            ({ denom }) => {
-              return denom === srcFeeAsset?.denom;
-            }
-          );
+          feeDenomPrices ??= srcChain.feeAssets.find(({ denom }) => {
+            return denom === srcFeeAsset?.denom;
+          });
```
toddkao, line 623: this looks redudant with lines 617-619

toddkao (approved):
I'm finding alot of the logic in `use-swap-widget` surprisingly difficult to follow, but as far as the refactor is concerned this looks good. The logic is definitely something I'll want to see if we can simplify in v2

packages/widget/src/hooks/use-swap-widget.ts
codingki: right, nice catch

codingki:
> I'm finding alot of the logic in `use-swap-widget` surprisingly difficult to follow, but as far as the refactor is concerned this looks good. The logic is definitely something I'll want to see if we can simplify in v2

I agree with you, it is difficult to follow. I'm avoiding to refactor this legacy code. Yes we need to simplify it in v2

PR #109, opened 2024-07-29
[FRE-892] Add MainButton Component

toddkao:
Screenshot 2024-07-30 at 12 17 04 PM

https://github.com/user-attachments/assets/ccefe746-4d3b-47c0-87b5-0ae722bb74f9


packages/widget-v2/src/components/MainButton.tsx
```diff
@@ -0,0 +1,165 @@
...
+          </Text>
+          {loadingTimeString && (
+            <StyledTimeRemaining align="center" justify="center">
+              <SmallText>{loadingTimeString}</SmallText>
+            </StyledTimeRemaining>
+          )}
+        </StyledOverlay>
+      </StyledMainButton>
+    );
+  }
```
toddkao, line 80: This loading animation is not perfect, not sure if we want to switch to using something better but I'm gonna leave it like this for now

packages/widget-v2/src/components/MainButton.tsx
```diff
@@ -0,0 +1,165 @@
...
+};
+
+export type MainButtonProps = {
+  label: string;
+  icon?: ICONS | 'none';
+  leftIcon?: ICONS | 'none';
+  backgroundColor?: string;
+  disabled?: boolean;
+  loading?: boolean;
+  loadingTimeString?: string;
```
codingki, line 37: add `onClick`

codingki: or you can extends button props and spread it in `StyledMainButton` props

codingki (changes requested):

packages/widget-v2/src/components/MainButton.tsx
```diff
@@ -0,0 +1,168 @@
...
+  const LeftIcon = iconMap[leftIcon];
+
+  if (loading) {
+    return (
+      <StyledMainButton
+        align="center"
+        justify="center"
+        backgroundColor={theme.backgroundColor}
+        loading={loading}
+      >
```
toddkao, line 64: I didn't add onClick handler on this because it seems like the "loading" buttons shouldn't be clickable

codingki: yeah onClick can be undefined so lgtm

codingki (approved):

PR #132, opened 2024-08-05
[FRE-856] fix(docs): example guide

codingki:


NotJeremyLiu (approved):
did a pass through, made a small commit for some wording differences, lgtm, ty sir!

We should probably wait until the client change goes in and then update this PR to use client terminology, approving now though

PR #137, opened 2024-08-06
Find last swapAction that meets criteria in makeChainIdsWithActions

toddkao:
new:
Screenshot 2024-08-06 at 3 08 10 PM

old:
Screenshot 2024-08-06 at 3 08 33 PM

### DIFF
--- a/packages/widget/src/ui/PreviewRoute/make-chain-ids-with-actions.ts
+++ b/packages/widget/src/ui/PreviewRoute/make-chain-ids-with-actions.ts
@@ -35,6 +35,7 @@ export const makeChainIDsWithAction = ({
     });
 
     const reversedActions = actions.reverse();
+    // Find the last swap action that matches conditions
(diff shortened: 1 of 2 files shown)


packages/widget/src/ui/PreviewRoute/make-chain-ids-with-actions.ts
```diff
@@ -34,7 +34,9 @@ export const makeChainIDsWithAction = ({
       );
     });
 
-    const swapAction = actions.find((action) => {
+    const reversedActions = actions.slice().reverse();
+    // Find the last swap action that matches conditions
+    const swapAction = reversedActions.find((action) => {
```
toddkao, line 39: Seems like the easiest way to do this, too bad findLast is still pretty new ([came out in like 2023](https://www.w3schools.com/js/js_2023.asp)) and lodash/iterating backwards manually in the array feels like overkill

PR #155, opened 2024-08-08
Create TokenAndChainSelector Flow

toddkao:
I didn't do the part where we show balances or the arrow (default chain) since we don't have that data yet.

I'd also like to add keyboard navigation but haven't done that yet either, will get around to that later
 
https://github.com/user-attachments/assets/37460050-507f-4539-aac9-133423b14e4c


packages/widget-v2/src/widget/TokenAndChainSelectorFlow/TokenAndChainSelectorFlow.tsx
```diff
@@ -0,0 +1,110 @@
...
+    }, [searchQuery, chainsContainingAsset]);
+
+    useEffect(() => {
+      if (!isPending && assets) {
+        const timer = setTimeout(() => {
+          setShowSkeleton(false);
+        }, 100);
+        return () => clearTimeout(timer);
+      }
+    }, [isPending, assets]);
```
toddkao, line 53: I added a skeleton/fallback on mount because even when the data exists, it takes a while for it to fully render which would normally cause a delay before the modal opens. This way we open the modal faster and swap out the real List afterwards and it doesn't feel as laggy.

packages/widget-v2/src/widget/TokenAndChainSelectorFlow/TokenAndChainSelectorFlow.tsx
```diff
@@ -0,0 +1,110 @@
...
+                <Skeleton key={index} />
+              ))}
+            </Column>
+          ) : (
+            <VirtualList
+              listItems={filteredChains ?? filteredAssets ?? []}
+              height={530}
+              itemHeight={70}
+              renderItem={renderItem}
+              itemKey={(item) => hashObject(item)}
```
codingki, line 93: can we use `${item.chainID}-${item.denom}`

codingki:
the perfomance not quite good, it's fast if have a filter 
https://github.com/user-attachments/assets/214da552-6144-4923-a77b-d8ca5087bbe0

packages/widget-v2/src/widget/TokenAndChainSelectorFlow/TokenAndChainSelectorFlow.tsx
toddkao: yeah I just noticed for some reason that still had duplicate keys, I'm guessing you filter them out?

toddkao: alright I'm gonna remove the hashObject fn

codingki (approved):

PR #161, opened 2024-08-12
Add chainIDsToAffiliates to SkipRouterOptions, Add chainIDsToAffiliates to SwapWidget props

toddkao:
Add chainIDsToAffiliates to SkipRouterOptions

manually verified that both cumulative_affiliate_bps is passed to /route
Screenshot 2024-08-12 at 2 46 39 PM

and affiliates is passed to /msgs 
Screenshot 2024-08-12 at 2 47 05 PM

Add test for validateChainIDsToAffiliates to validate chainIDsToAffiliates data is not malformed

Screenshot 2024-08-12 at 2 51 54 PM

### DIFF
--- a/.changeset/odd-cycles-greet.md
+++ b/.changeset/odd-cycles-greet.md
@@ -1,0 +1,6 @@
+---
+'@skip-go/client': minor
+'@skip-go/widget': minor
+---
+
+Add chainIDsToAffiliates to SkipRouterOptions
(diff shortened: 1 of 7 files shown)


.changeset/odd-cycles-greet.md
```diff
@@ -0,0 +1,6 @@
+---
+'@skip-go/client': minor
+'@skip-go/widget': minor
```
codingki, line 3: this will cause a failed workflow, I suggest to bump client first

codingki (approved):
lgtm, except the changeset

PR #165, opened 2024-08-13
Rename skip router to skip client

codingki:


packages/client/src/__test__/client.test.ts
```diff
@@ -1,7 +1,7 @@
 import { rest } from 'msw';
 import { setupServer } from 'msw/node';
 
-import { SKIP_API_URL, SkipRouter } from '../client';
+import { SKIP_API_URL, SkipClient, SkipRouter } from '../client';
```
toddkao, line 4: forgot to remove SkipRouter import here

toddkao (approved):

packages/client/src/__test__/client.test.ts
codingki: on purpose, for testing the deprecated one

PR #170, opened 2024-08-14
Add SwapExecutionFlow

toddkao:
https://github.com/user-attachments/assets/c0e78c87-df18-4e36-b3b1-d29dec74bd7e


packages/widget-v2/src/utils/misc.ts
```diff
@@ -1,6 +0,0 @@
-import crypto from 'crypto';
```
codingki, line 1: can we check if we use `crypto` package anywhere, if not let's remove it from our deps

codingki (approved):
left a couple comments, but lgtm

packages/widget-v2/src/utils/misc.ts
toddkao: @codingki we don't using crypto anywhere else, but also it's not an explicit dep anywhere in our package.json it's a core node_module which is being polyfilled in our vite.config.ts

PR #173, opened 2024-08-16
Fix penumbra and add penumbra to the widget

codingki:
Previously prax wallet is injected outside the widget, now it's included in the widget


toddkao (approved):
just gotta remove the console.logs

PR #181, opened 2024-08-20
Add Router, errorAtom, renderLogic

toddkao:


packages/widget-v2/tsconfig.app.json
```diff
@@ -26,7 +26,7 @@
     /* Aliases */
     "baseUrl": ".",
     "paths": {
-      "@*": ["src/*"]
+      "@/*": ["src/*"]
```
toddkao, line 29: This fixes auto imports / ide from suggesting `@components` instead of `@/components` which isn't valid with this config

packages/widget-v2/src/widget/Router.tsx
```diff
@@ -0,0 +1,31 @@
...
+        </ErrorBoundary>
+      );
+    case Routes.SwapExecutionPage:
+      return (
+        <ErrorBoundary fallback={null} onError={(error) => setError(error)}>
+          <SwapExecutionPage />
+        </ErrorBoundary>
+      );
+  }
+};
```
toddkao, line 31: I tested manually throwing an error within SwapPage and it successfully sets errorAtom state and renders ErrorPage

toddkao: I also tested and can confirm that these error boundaries **will not** catch any errors thrown within a modal, we will need to handles those seperately

packages/widget-v2/src/pages/ErrorPage/ErrorPage.tsx
```diff
@@ -0,0 +1,13 @@
...
+export const ErrorPage = () => {
+  const resetError = useResetAtom(errorAtom);
+
+  return (
+    <div>
+      error page
+      <button onClick={() => resetError()}>reset</button>
+    </div>
+  );
+};
```
toddkao, line 13: this is temp, i created a ticket to actually create the ErrorPage properly

packages/widget-v2/src/stories/SwapExecutionPage.stories.tsx
```diff
@@ -57,7 +56,5 @@ type Story = StoryObj<typeof meta>;
 export default meta;
 
 export const SwapExecutionPageExample: Story = {
-  args: {
-    operations: operations as Operation[],
-  },
+  args: {},
```
codingki, line 59: any reason why you move it inside?

codingki (approved):
lgtm

packages/widget-v2/src/stories/SwapExecutionPage.stories.tsx
toddkao: @codingki yeah I figure that eventually this data will exist in an atom, so we won't be passing it into the component as a prop, so I decided to just move it inside for now (the usage of the component is more similar to how it'll be when we have the data in an atom)

PR #186, opened 2024-08-21
[API-3285] add expiration and memo fields to CosmosTx object

dhfang:


codingki (approved):

dhfang:
the design changed and these fields are no more

PR #188, opened 2024-08-26
feat: inject wallet from parent

codingki:
to test /inject-wallet


toddkao (approved):
like we discussed, the way we currently do it works for now, but I think we should try to improve the DX in V2 if possible

consuming the providers if our component is wrapped within one would be ideal

codingki:
Sei team approved

PR #191, opened 2024-08-27
[FRE-951, FRE-952]refactor(widget-v2): remove chain registry usage for chain data

codingki:
chain registry is only for cosmos chains, better use chain data from skip api/client.

### DIFF
--- a/packages/widget-v2/src/state/chains.ts
+++ b/packages/widget-v2/src/state/chains.ts
@@ -1,0 +1,43 @@
+
+import { Chain, AssetList } from "@chain-registry/types";
+import {
+  chains as chainsChainRegistry,
+  assets as assetsChainRegistry,
+} from "chain-registry";
+import {
+  chains as chainsInitiaRegistry,
+  assets as assetsInitiaRegistry,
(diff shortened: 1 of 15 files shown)


packages/widget-v2/src/state/chains.ts
```diff
@@ -0,0 +1,43 @@
...
+  const chainName = chainIdToName(chainId);
+  const assetsFoundForChain = assets.find(
+    (a) => a.chain_name === chainName
+  )?.assets;
+
+  if (!assetsFoundForChain) {
+    throw new Error(`chain '${chainId}' does not exist in assetsRecord`);
+  }
+  return assetsFoundForChain;
+}
```
toddkao, line 43: @codingki is it possible for us to get rid of the dependency on chain-registry / @initia/initia-registry completely at some point? i noticed they are pretty big dependencies

toddkao (approved):
👍  good refactor

packages/widget-v2/src/state/chains.ts
codingki: Nope, We will need it for wallet connecting

PR #193, opened 2024-08-29
[FRE-937, FRE-950, FRE-953, FRE-954, FRE-957] feat(widget-v2): add skip client

codingki:


.changeset/forty-crews-relax.md
codingki, line 1: I forgot to add changeset in previous PR #188, let's bump it with this PR

toddkao (approved):

PR #206, opened 2024-08-30
[FRE-960] feat(widget-v2): swap page route logic

codingki:


packages/widget-v2/src/components/AssetChainInput.tsx
```diff
@@ -48,8 +50,63 @@ export const AssetChainInput = ({
...
+                  value = value.minus(1);
+                }
+              }
+              if (value.isNegative()) {
+                value = new BigNumber(0);
+              }
+              onChangeValue(value.toString());
+            }
+          }
+          }
```
toddkao, line 109: can we refactor onKeyDown to use switch cases? something like this
```
if (!onChangeValue) return;

  let value = new BigNumber(
    formatNumberWithoutCommas(event.currentTarget.value) || "0"
  );
  switch (event.key) {
    case "Escape":
      if (
        event.currentTarget.selectionStart ===
        event.currentTarget.selectionEnd
      ) {
        event.currentTarget.select();
      }
      return;

    case "ArrowUp":
      event.preventDefault();

      if (event.shiftKey) {
        value = value.plus(10);
      } else if (event.altKey || event.ctrlKey || event.metaKey) {
        value = value.plus(0.1);
      } else {
        value = value.plus(1);
      }
    case "ArrowDown":
      event.preventDefault();

      if (event.shiftKey) {
        value = value.minus(10);
      } else if (event.altKey || event.ctrlKey || event.metaKey) {
        value = value.minus(0.1);
      } else {
        value = value.minus(1);
      }
      
      break;

    default:
      break;
  }
  if (value.isNegative()) {
    value = new BigNumber(0);
  }
  onChangeValue(value.toString());
```

toddkao: we could also combine ArrowUp and ArrowDown code and just determine which function / plus or minus based on the direction if we wanted to

codingki: sure

toddkao:
```
    case "ArrowUp":
    case "ArrowDown":
      event.preventDefault();
      const incrementOrDecrementValue = event.key === 'ArrowUp' ? value.plus : value.minus;

      if (event.shiftKey) {
        value = incrementOrDecrementValue(10);
      } else if (event.altKey || event.ctrlKey || event.metaKey) {
        value = incrementOrDecrementValue(0.1);
      } else {
        value = incrementOrDecrementValue(1);
      }

      break;
```

packages/widget-v2/src/pages/SwapPage/SwapPage.tsx
```diff
@@ -23,19 +32,28 @@ export const SwapPage = () => {
   const [{ data: assets }] = useAtom(skipAssetsAtom);
   const [{ data: chains }] = useAtom(skipChainsAtom);
   const [destinationAsset, setDestinationAsset] = useAtom(destinationAssetAtom);
-
+  const [_, setSwapDirection] = useAtom(swapDirectionAtom);
```
toddkao, line 35: I'm just learning this myself, we should use [useSetAtom](https://jotai.org/docs/core/use-atom#usesetatom)if we dont need to actually get the value

toddkao (approved):

PR #207, opened 2024-09-02
widget v2: invert swap logic

codingki:


packages/widget-v2/src/pages/SwapPage/SwapPage.tsx
```diff
@@ -168,9 +168,9 @@ export const SwapPage = () => {
             }}
           />
         </Column>
-        {!isRouteLoading && <MainButton label="Connect Wallet" icon={ICONS.plus} />}
+        {!isRouteLoading && !isRouteError && <MainButton label="Connect Wallet" icon={ICONS.plus} />}
         {isRouteLoading && <MainButton label="Finding Best Route..." loading={true} />}
-
+        {isRouteError && !isRouteLoading && <MainButton label={routeError.message} disabled={true} />}
```
toddkao, line 173: can we move this render logic to a useMemo now that it's growing in number of cases?

the logic should look something like this

if (isRouteError) {
return <MainButton label={err}
}
if (isRouteLoading {
return ''
}
return <MainButton label="connectWallet"

codingki: this logic is not complete, we can merge and improve this later

PR #208, opened 2024-09-03
Fix filter + defaultRoute bug

toddkao:
Fix filter bug

### DIFF
--- a/packages/widget/src/hooks/use-swap-widget.ts
+++ b/packages/widget/src/hooks/use-swap-widget.ts
@@ -70,9 +70,6 @@ export function useSwapWidget(persistSwapWidgetState = true) {
   const destinationChainIDs = filter?.destination
     ? Object.keys(filter.destination)
     : undefined;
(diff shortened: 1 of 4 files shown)


packages/widget/src/hooks/use-swap-widget.ts
```diff
@@ -70,24 +70,14 @@ export function useSwapWidget(persistSwapWidgetState = true) {
...
     getAsset,
   } = useAssets();
-  const { data: chains } = useChains({
-    select: (chains) => {
-      if (allChainIDs.length === 0) return chains;
-      return chains?.filter(({ chainID }) => {
-        return allChainIDs.includes(chainID);
-      });
-    },
-  });
```
toddkao, line 90: if either filter.source or filter.destination is undefined we will effectively filter all of those chains out, and it's not super feasible to simply include "all of the chainIDS" when filter.source/filter.destination is undefined, let's simplify this to simply not select any chains and always return all of the chains

PR #215, opened 2024-09-03
Widget v2/wallet provider

codingki:


packages/widget-v2/src/constants/graz.ts
```diff
@@ -0,0 +1,22 @@
...
+import { ChainInfo } from "@keplr-wallet/types";
+
+
+
+export const mainnetChains = _mainnetChains.map((chain) => {
+  try {
+    return chainRegistryChainToKeplr(chain, _mainnetAssets)
+  } catch (_error) {
+    return false
+  }
```
codingki, line 13: this is due a error thrown by `chainRegistryChainToKeplr` function, so we need a try catch

toddkao (approved):

PR #218, opened 2024-09-05
Create wallets hook

codingki:


packages/widget-v2/src/state/wallets.ts
```diff
@@ -0,0 +1,39 @@
...
+}
+
+type WalletState = {
+  walletName: string;
+  chainType: string;
+}
+
+export const evmWaletAtom = atom<WalletState>()
+export const cosmosWaletAtom = atom<WalletState>()
+export const svmWaletAtom = atom<WalletState>()
```
toddkao, line 31: "Walet" typo

toddkao (approved):
looks good besides typo

packages/widget-v2/src/hooks/useCreateCosmosWallets.tsx
```diff
@@ -0,0 +1,130 @@
...
+        connect: async () => {
+          if (wallet === currentWallet) return
+          try {
+            if (!chainInfo) throw new Error(`connect: Chain info not found for chainID: ${chainID}`)
+            await suggestAndConnectAsync({ walletType: wallet, chainInfo })
+            setCosmosWallet({ walletName: wallet, chainType: "cosmos" })
+            // TODO: onWalletConnected
+          } catch (error) {
+            console.error(error)
+            throw error
```
toddkao, line 101: we should test what happens if an error occurs and we throw the error here

packages/widget-v2/src/hooks/useCreateEvmWallets.tsx
```diff
@@ -0,0 +1,94 @@
...
+        },
+        connect: async () => {
+          if (connector.id === currentEvmConnector?.id) return;
+          try {
+            await connectAsync({ connector, chainId: Number(chainID) });
+            setEvmWallet({ walletName: connector.id, chainType: "evm" });
+            // TODO: onWalletConnected
+          } catch (error) {
+            console.error(error);
+            throw error;
```
toddkao, line 49: same with this one, we should test what happens if this error is thrown

packages/widget-v2/src/state/chains.ts
```diff
@@ -8,16 +8,39 @@ import {
...
 ] as AssetList[];
 
+
+export const chainInfos = chains.map((chain) => {
+  try {
+    return chainRegistryChainToKeplr(chain, assets)
+  } catch (_error) {
+    return false
+  }
+}).filter(Boolean) as ChainInfo[]
```
toddkao, line 31: I don't like this at all... I think we should refactor this

toddkao: also im a bit confused about the try catch `chainRegistryChainToKeplr` doesn't look like it returns a promise

toddkao: I'll test it myself and see

toddkao: @codingki let's refactor to this: we can use the getChainInfo function here and simply filter out undefineds
```ts
export const getChainInfo = (chainID: string) => {
  const chain = chains.find((chain) => chain.chain_id === chainID);
  if (!chain) return undefined;
  try {
    return chainRegistryChainToKeplr(chain, assets);
  } catch (_error) {
    return undefined;
  }
};

export const allChainInfo = chains
  .map((chain) => getChainInfo(chain.chain_id))
  .filter((info) => info);
```

toddkao: tbh `.filter(Boolean)` and `.filter((chainInfo) => chainInfo))` are exactly the same but for some reason i hate the first one it's unintuitive

packages/widget-v2/src/hooks/useCreateEvmWallets.tsx
codingki: yeah I'm still figuring out how we going to display these errors without toast

packages/widget-v2/src/state/chains.ts
codingki: so `chainRegistryChainToKeplr` is not a promise but it can throw an error, so we need to wrap it with try catch. I can't find which chain is causing the error, but you can try converting `chains` from `chain-registry` it will show you the client side error. with try catch wrapper we can just not returning the chain if it has a convertion error

codingki: yeah `undefined` is a good suggestion

PR #219, opened 2024-09-05
[FRE-966] Refactor/finality time using route repsonse

codingki:
image


packages/widget/src/ui/HistoryDialog/HistoryList.tsx
```diff
@@ -238,11 +235,9 @@ export const Item = forwardRef<HTMLDivElement, ItemProps>(function Item(
             </DescriptionList.Row>
           ))}
           <DescriptionList.Row>
-            <DescriptionList.Dt>Completion Time</DescriptionList.Dt>
+            <DescriptionList.Dt>Estimation Completion Time</DescriptionList.Dt>
```
toddkao, line 238: is this the right text? should it be "Estimated Completion time"?

toddkao (approved):

packages/widget/src/ui/HistoryDialog/HistoryList.tsx
codingki: yeah because we don't know the real completion time

toddkao: @codingki I'm asking if it should be "Estimated" instead of "Estimation", not a big deal but Estimation Completion Time sounds weird to me

codingki: hmm you right, I'll fix the typo before release

PR #223, opened 2024-09-09
wallet selector

codingki:


packages/widget-v2/src/widget/wallet/CosmosProvider.tsx
```diff
@@ -13,9 +13,6 @@ export const CosmosProvider: React.FC<CosmosProviderProps> = ({
     <GrazProvider grazOptions={{
       chains: mainnetChains,
       autoReconnect: true,
-      onNotFound: () => {
-        console.warn("wallet not found")
-      },
```
toddkao, line 18: I think we should keep this around, maybe we can handle it better in the future and but this way we won't forget this case can happen

packages/widget-v2/src/pages/SwapPage/SwapPage.tsx
```diff
@@ -36,6 +37,7 @@ export const SwapPage = () => {
   const { isLoading: isRouteLoading, isError: isRouteError, error: routeError } = useAtomValue(skipRouteAtom);
   const swapFlowSettings = useModal(SwapPageSettings);
   const tokenAndChainSelectorFlow = useModal(TokenAndChainSelectorModal);
+  const selectWalletFlow = useModal(WalletSelectorModal);
```
toddkao, line 40: nit: rename these vars to "swapPageSettings, tokenAndChainSelectorModal, walletSelectorModal", let's get rid of the word "flow" when we refactor

packages/widget-v2/src/components/RenderWalletList.tsx
```diff
@@ -83,12 +82,12 @@ export const RenderWalletList = ({
                   height={35}
                   width={35}
                   style={{ objectFit: "cover" }}
-                  src={imageUrl}
-                  alt={`${name} logo`}
+                  src={imageUrl ?? "https"}
```
toddkao, line 85: I don't get the "https" fallback, it'll show an empty image just like an empty string right? no big deal just seems random

packages/widget-v2/src/components/RenderWalletList.tsx
```diff
@@ -54,71 +53,130 @@ export const RenderWalletListHeader = ({
...
+          listItems={walletList}
+          height={height}
+          itemHeight={ITEM_HEIGHT + ITEM_GAP}
+          renderItem={renderItem}
+          itemKey={(item) => item.walletName}
+        />;
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [connectMutation, height, walletList]);
+
```
toddkao, line 161: i think in this case we just use if statements instead of switch

```ts
if (connectMutation.isError) {
  return ...
} 
if (connectMutation.isPending) {
  return ...
}
return ()
```

packages/widget-v2/src/components/RenderWalletList.tsx
```diff
@@ -54,71 +53,132 @@ export const RenderWalletListHeader = ({
...
+              <img
+                style={{ objectFit: "cover" }}
+                src={connectMutation.variables?.walletInfo.logo}
+                alt={`${connectMutation.variables?.walletPrettyName} logo`} />
+            </StyledAnimatedBorder>
+            <Text color={theme.primary.text.lowContrast}>Connecting to {connectMutation.variables?.walletPrettyName}</Text>
+          </StyledLoadingContainer>
+        </StyledInnerContainer>
+      );
+    }
```
toddkao, line 151: @codingki these two are very similar, is there a state that is true for both of them? I think we can combine them and then just use different txState and text for the "connecting to" and "failed to connect to", since everything else is exactly the same

codingki: sure

packages/widget-v2/src/hooks/useWalletList.tsx
```diff
@@ -0,0 +1,65 @@
...
+        }
+      case chainType === "cosmos":
+        return createCosmosWallets(chainID);
+      case chainType === "evm":
+        return createEvmWallets(chainID);
+      case chainType === "svm":
+        return createSolanaWallets();
+      default:
+        return [];
+    }
```
toddkao, line 61: I think we should do this:

```ts
let walletType = chainType;
const isSei = chainID === "pacific-1";
if (isSei) {
  walletType = "sei";
}
switch(walletType)
  case "cosmos"
  case "sei"
  case "evm"
  case "svm"
  default:
```

packages/widget-v2/src/components/RenderWalletList.tsx
```diff
@@ -54,71 +53,132 @@ export const RenderWalletListHeader = ({
...
+    return <VirtualList
+      listItems={walletList}
+      height={height}
+      itemHeight={ITEM_HEIGHT + ITEM_GAP}
+      renderItem={renderItem}
+      itemKey={(item) => item.walletName}
+    />;
+
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [connectMutation, height, walletList]);
```
toddkao, line 162: @codingki why are we disabling exhaustive-deps?

toddkao: I don't think we should ever disable this warning btw

toddkao: I suggest you just move the renderItem to within this useMemo

packages/widget-v2/src/hooks/useCreateEvmWallets.tsx
```diff
@@ -1,24 +1,31 @@
...
-
     const wallets: MinimalWallet[] = [];
     for (const connector of connectors) {
+      if (
+        wallets.findIndex((wallet) => wallet.walletName === connector.id) !==
+        -1
+      ) {
+        continue;
+      }
+
```
toddkao, line 28: I think it'd be easier read if we simply wrap everything with

```ts
const walletAlreadyConnected = wallets.find((wallet) => wallet.walletName === connector.id);
 if (!walletAlreadyConnected) {
  ...
 }
 ```

File: packages/widget-v2/src/components/RenderWalletList.tsx [hunk 9]
```diff
@@ -54,71 +53,111 @@ export const RenderWalletListHeader = ({
...
+        rightContent={rightContent?.()}
+      />
+    );
+  }, [connectMutation]);
 
   const height = useMemo(() => {
     return Math.min(530, walletList.length * (ITEM_HEIGHT + ITEM_GAP));
   }, [walletList]);
 
+  const container = useMemo(() => {
```
toddkao, line 108: let's rename this to something more descriptive like:

```ts
const renderWalletListOrWalletConnectionStatus
```

toddkao (approved):

PR #232, opened 2024-09-12
Add debouncing route request

toddkao:
Add debouncing route request using atomWithDebounce
https://jotai.org/docs/recipes/atom-with-debounce

Add eslint rule to enforce semicolons

### DIFF
--- a/packages/widget-v2/src/utils/atomWithDebounce.ts
+++ b/packages/widget-v2/src/utils/atomWithDebounce.ts
@@ -1,0 +1,67 @@
+import { atom } from "jotai"
+import { SetStateAction } from "react"
+
+export function atomWithDebounce<T>(
+  initialValue: T,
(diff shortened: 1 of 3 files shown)


packages/widget-v2/src/utils/atomWithDebounce.ts
```diff
@@ -0,0 +1,67 @@
...
+    set(isDebouncingAtom, false);
+  });
+
+  return {
+    currentValueAtom: atom((get) => get(_currentValueAtom)),
+    isDebouncingAtom,
+    clearTimeoutAtom,
+    debouncedValueAtom,
+  };
+}
```
toddkao, line 67: this entire file is copied from https://jotai.org/docs/recipes/atom-with-debounce

codingki (approved):

PR #233, opened 2024-09-12
[FRE-967] Fix validate gas balance checking

codingki:


packages/client/src/client.ts
```diff
@@ -272,10 +272,10 @@ export class SkipClient {
       gasAmountMultiplier = DEFAULT_GAS_MULTIPLIER,
       getFallbackGasAmount,
     } = options;
-    let gasTokenUsed: Coin | undefined;
+    let gasTokensUsed: Record<string, Coin> | undefined;
     if (validateGasBalance) {
       // check balances on chains where a tx is initiated
-      gasTokenUsed = await this.validateGasBalances({
+      gasTokensUsed = await this.validateGasBalances({
```
toddkao, line 278: does `this.validateGasBalances` return a record<string, Coin> of gasTokenUsed?

codingki: yes

toddkao: @codingki it's a little confusing having `gasTokenUsed` and `gasTokensUsed`, maybe change one of them to be called GasTokenArray? also is gasTokenUsed an array or record? `gasTokensUsed?.[i]` being used in a loop suggests its an array

codingki: it's a record. we can use a record like that

packages/client/src/client.ts
```diff
@@ -1778,6 +1776,8 @@ export class SkipClient {
     gasAmountMultiplier?: number;
     getFallbackGasAmount?: clientTypes.GetFallbackGasAmount;
   }) {
+    // tx index -> gas token used
+    let gasTokenUsed: Record<number, Coin> = {}
```
toddkao, line 1780: @codingki let's update the name of this one to gasTokenRecord as well, and should they both be Record<number, Coin> ?

codingki: updated, yes it should be a record. because not all tx is a `cosmosTx` so it won't be always sorted, can be random index

toddkao: @codingki I understand we want it to be a record I just mean the other one is `Record<string, Coin>` instead of `Record<number, Coin>`

packages/client/src/client.ts
```diff
@@ -272,10 +272,10 @@ export class SkipClient {
       gasAmountMultiplier = DEFAULT_GAS_MULTIPLIER,
       getFallbackGasAmount,
     } = options;
-    let gasTokenUsed: Coin | undefined;
+    let gasTokenRecord: Record<string, Coin> | undefined;
```
codingki, line 275:
```suggestion
    let gasTokenRecord: Record<number, Coin> | undefined;
```

toddkao (approved):

PR #239, opened 2024-09-13
Fix debouncedRouteRequestEffect

toddkao:
currently the effect is triggering too often, we need to only set the `debouncedSkipRouteRequestAtom` when it differs from `newRouteRequest` using `lodash.isEqual`


codingki (approved):

packages/widget-v2/package.json
```diff
@@ -34,6 +34,7 @@
     "@storybook/react-vite": "^8.2.6",
     "@storybook/test": "^8.2.6",
     "@types/eslint__js": "^8.42.3",
+    "@types/lodash.isequal": "^4.5.8",
```
toddkao, line 37: i can remove this now

codingki (approved):

PR #242, opened 2024-09-13
Add isWaitingToUpdateInputValue + animation to AssetChainInput, Add priceChangePercentage

toddkao:
1. Add Red/Green arrow and price change % on hover
2. Add isWaitingToUpdateInputValue + animation
3. Add invertSwapAtom to prevent unnecessary route request
4. Add spinning animation when clicking on invertSwap

https://github.com/user-attachments/assets/97c0967e-0be3-44c6-b396-07b6b6798d31


packages/widget-v2/src/state/swapPage.ts
```diff
@@ -11,39 +11,81 @@ export type AssetAtom = Partial<ClientAsset> & {
...
+    set(sourceAssetAmountAtom, destinationAsset?.amount);
+  }
+
+  set(destinationAssetAtom, sourceAsset);
+  if (sourceAsset?.amount) {
+    set(destinationAssetAmountAtom, sourceAsset?.amount);
+  }
+
+  set(isInvertingSwapAtom, false);
+});
```
toddkao, line 88: isInvertingSwapAtom helps to reduce the amount of route requests made when swapping between source and destination assets to 1

codingki: it still need to fetch the route

toddkao: @codingki yeah it still needs to fetch the route but before we were fetching extra times

PR #245, opened 2024-09-16
[FRE-982] fix: zero appeared in tx preview

codingki:


toddkao (approved):

toddkao (commented):
this is one of the reasons why one of the companies I worked at suggested we always rendering with ternary 😆 but at the time react would also crash if returning `undefined` in the render so it was much worse

```ts
route.estimatedRouteDurationSeconds ? <div></div> : undefined
```

PR #248, opened 2024-09-16
Add missing dependency

toddkao:


codingki:
uh I don't think we need this dependancy anymore

PR #251, opened 2024-09-16
[FRE-983] widget v2: connected wallet state

codingki:
https://github.com/user-attachments/assets/24bd3319-7d3e-4e28-8031-61c9aa470306


examples/nextjs/pages/v2.tsx
```diff
@@ -0,0 +1,39 @@
...
+          height: '40px',
+        }}
+      >
+        Purge Settings
+      </button>
+    </div>
+  );
+};
+
+export default Home;
```
toddkao, line 39: we cant commit this cuz the v2 build wont exist in vercel env

packages/widget-v2/src/components/Button.tsx
```diff
@@ -28,6 +28,37 @@ export const GhostButton = styled(SmallText).attrs({
...
+        : theme.primary.ghostButtonHover};
+        color: ${theme.primary.text.normal};
+        cursor: pointer;
+      `};
+  }
+  padding: 10px 13px;
+  border-radius: 90px;
+  ${flexProps};
+`;
+
```
toddkao, line 61: I would suggest:

1. Move this to SwapPage.tsx since this isn't a generic button
2. TransparentButton should be based on GhostButton like this and it will inherit all of GhostButton styles so you don't need to duplicate them

```ts
const TransparentButton = styled(GhostButton)``
```

examples/nextjs/pages/v2.tsx
codingki: oh this mistakenly commited

toddkao (approved):

PR #254, opened 2024-09-17
Add SwapExecutionPage

toddkao:
Store route in swapExecutionStateAtom
use swapExecutionStateAtom to render SwapExecutionPage


packages/widget/web-component/package.json
codingki, line 1: is this meant to be deleted?

packages/widget-v2/src/modals/ManualAddressModal/ManualAddressModal.tsx
```diff
@@ -20,7 +20,7 @@ export const ManualAddressModal = createModal((modalProps: ModalProps) => {
   const { theme } = modalProps;
   const modal = useModal();
   const [destinationAsset] = useAtom(destinationAssetAtom);
-  const [, setDestinationWallet] = useAtom(destinationWalletAtom);
+  const [, _setDestinationWallet] = useAtom(destinationWalletAtom);
```
codingki, line 23: we can remove this line

codingki (approved):

codingki:
there are some eslint error u need to solve

packages/widget/web-component/package.json
toddkao: oh idk why its deleted, ill revert

toddkao: to be honest it doesn't really matter because this file is autogenerated when you run the publish script but yeah ill revert to reduce the diff size

PR #259, opened 2024-09-18
Add using skipClient.balances

toddkao:
Add using skipClient.balances 
Use devApiUrl
Fix invertSwapAtom logic
Add EVM warning 

https://github.com/user-attachments/assets/2def16e2-6691-4c44-89f1-33ed54bf9bf0

Ensure that debouncing still works when inverting source/dest

https://github.com/user-attachments/assets/ba37bbaa-4ead-4104-8b21-0d028fc0e07e

Screenshot 2024-09-19 at 3 37 37 PM


packages/widget-v2/src/pages/SwapPage/useSetMaxAmount.ts
```diff
@@ -0,0 +1,75 @@
...
+import { convertHumanReadableAmountToCryptoAmount, convertTokenAmountToHumanReadableAmount } from "@/utils/crypto";
+import { useGetAssetDetails } from "@/hooks/useGetAssetDetails";
+import { sourceAssetAmountAtom, sourceAssetAtom } from "@/state/swapPage";
+import { useAtom, useSetAtom } from "jotai";
+import { skipChainsAtom } from "@/state/skipClient";
+import { useSourceBalance } from "./useSourceBalance";
+import { BigNumber } from "bignumber.js";
+
+
+const ETH_GAS_FEE = 0.1;
```
codingki, line 11: this should be 0.01 or 10 * 10 ** 16

codingki (changes requested):
can we limit digits after comma to 4 for the balance

packages/widget-v2/src/pages/SwapPage/useSetMaxAmount.ts
toddkao: I think it's a little easier to understand if we keep the value in terms of how much ETH and then convert it to the tokenAmount using a util like this:

`convertHumanReadableAmountToCryptoAmount(ETH_GAS_FEE)`

but I do need to pass the correct decimals which i forgot to do

packages/widget-v2/src/pages/SwapPage/useSetMaxAmount.ts
```diff
@@ -21,13 +21,18 @@ export const useGasFeeTokenAmount = () => {
     chainId: sourceAsset?.chainID,
   });
 
+  const ethDetails = useGetAssetDetails({
+    assetDenom: "ethereum-native",
+    chainId: "1"
```
codingki, line 26: you need to specify the denom for different chainId if you do it like this. It's easier to not do this because in evm it's not always eth for the fee asset

packages/widget-v2/src/pages/SwapPage/useSetMaxAmount.ts
```diff
@@ -0,0 +1,75 @@
...
+    chainId: sourceAsset?.chainID,
+  });
+
+  const feeAsset = chains?.find(chain => chain.chainID === sourceAsset?.chainID)?.feeAssets?.[0];
+
+  const chainType = sourceDetails?.chain?.chainType;
+
+  switch (chainType) {
+    case "evm":
+      return Number(convertHumanReadableAmountToCryptoAmount(ETH_GAS_FEE));
```
codingki, line 30: you need to pass decimals from source details, we are not sure every evm native asset is 18 decimals

toddkao: @codingki gotcha so the fee for evm is always in the source asset

codingki (approved):

PR #263, opened 2024-09-19
Add Max. slippage tooltip and update detailed row operation tooltip

toddkao:
https://github.com/user-attachments/assets/a80bc25e-ae23-4912-b4b4-0d87e2953bd0


packages/widget-v2/src/pages/SwapPage/SwapDetailModal.tsx
```diff
@@ -200,5 +234,20 @@ const SwapDetailText = styled(Row).attrs({
   as: SmallText,
   normalTextColor: true,
 })`
+  position: relative;
   letter-spacing: 0.26px;
 `;
+
+const Tooltip = styled(SmallText).attrs({
```
codingki, line 241: this might be useful in the future I think we can extract it to component

codingki (approved):

PR #265, opened 2024-09-19
widget-v2: set required chain addresses

codingki:


packages/widget-v2/src/pages/SwapExecutionPage/SwapExecutionPageRouteDetailedRow.tsx
```diff
@@ -52,38 +97,57 @@ export const SwapExecutionPageRouteDetailedRow = ({
...
+              {source.image && <img src={source.image} style={{
+                height: "100%"
+              }} />}
+              <StyledWalletAddress monospace>{`${source.address.slice(0, 9)}…${source.address.slice(-5)}`}</StyledWalletAddress>
+            </StyledButton>
           )}
         </Row>
-        <StyledWalletAddress>cosmos17...zha0v</StyledWalletAddress>
-      </Row>
+        {isSignRequired && <SmallText color={theme.warning.text}>Sign required</SmallText>}
```
toddkao, line 128: nit: text should be "Signature required" like in figma

toddkao (approved):

packages/widget-v2/src/pages/SwapExecutionPage/SwapExecutionPageRouteDetailedRow.tsx
```diff
@@ -52,38 +97,57 @@ export const SwapExecutionPageRouteDetailedRow = ({
...
-          {explorerLink && (
-            <Button onClick={() => window.open(explorerLink, "_blank")}>
-              <SmallText>
-                <ChainIcon />
-              </SmallText>
-            </Button>
+      <Column style={{
+        flex: 1,
+        justifyContent: "space-between"
+      }}>
```
toddkao, line 103: nit: use the justify prop instead of style for justifyContent

packages/widget-v2/src/pages/SwapExecutionPage/SwapExecutionPage.tsx
```diff
@@ -51,8 +53,43 @@ export const SwapExecutionPage = () => {
...
+
+    if (!isPending) {
+      if (allAddressesSet) {
+        return SwapExecutionState.ready;
+      }
+      if (!lastChainAddress) {
+        return SwapExecutionState.destinationAddressUnset;
+      }
+      return SwapExecutionState.recoveryAddressUnset;
+    }
```
toddkao, line 76: do we need to wrap these in !isPending? can we do something like this?

if (!isPending) return '';
if (allAddressesSet) {
  return SwapExecutionState.ready;
}
if (!lastChainAddress) {
  return SwapExecutionState.destinationAddressUnset;
}
return SwapExecutionState.recoveryAddressUnset;

codingki: it's easier to read if we wrap it in !isPending, because those state only happens before submit

PR #266, opened 2024-09-20
Feat/client solana support

ericHgorski:


packages/client/src/client.ts
```diff
@@ -118,6 +118,10 @@ export class SkipClient {
     this.getEVMSigner = options.getEVMSigner;
     this.getSVMSigner = options.getSVMSigner;
 
+    if (!Object.keys(this.endpointOptions).length) {
+      console.warn('Warning: You are using unreliable public endpoints. We strongly reccomend overriding them via endpointOptions for use beyond development settings.');
+    }
+
```
codingki, line 124: We shouldn't do this here, I think it's better to have the warning in `getRpcEndpointForChain`. if they not providing any rpc endpoint we output the warning.

codingki (changes requested):

packages/client/src/client.ts
codingki: https://github.com/skip-mev/skip-go/blob/950e8ee3865ad21fb60e68381be170e8d4bbe9eb/packages/client/src/client.ts#L1487

codingki: we can add it in that line

ericHgorski: fixed!

codingki (approved):
please do `npx changeset` and commit push before merging this.

PR #268, opened 2024-09-23
Add nuxtjs example

toddkao:
add nuxtjs to examples
use plugin to import from unpkg
add skip-widget to custom element in nuxtjs.config


ericHgorski (approved):
LGTM!

PR #269, opened 2024-09-24
Add handling transaction history

toddkao:
transaction history


toddkao:
> We shouldn't store the `status`. We should store the tx hashes every transactions. I suggest to create a custom hook to check the status. We should show status from the api, not storage.

@codingki 
We do already store the tx hashes (and other details of the transaction)
I was thinking about this a bit yesterday, here's what I'm thinking. We store the status for the transaction, but if there are any transactions in the history that are not completed (in some pending state) we can use the API to fetch the status in the background and then once they are completed, update the status. The logic is that once a transactions has completed, whether it did so successfully or failed the status is basically static and doesn't change anymore, but if pending we can use the API to fetch changes to the status.

codingki:
> > We shouldn't store the `status`. We should store the tx hashes every transactions. I suggest to create a custom hook to check the status. We should show status from the api, not storage.
> 
> @codingki We do already store the tx hashes (and other details of the transaction) I was thinking about this a bit yesterday, here's what I'm thinking. We store the status for the transaction, but if there are any transactions in the history that are not completed (in some pending state) we can use the API to fetch the status in the background and then once they are completed, update the status. The logic is that once a transactions has completed, whether it did so successfully or failed the status is basically static and doesn't change anymore, but if pending we can use the API to fetch changes to the status.

@toddkao  right, that works. so only if the status pending we fetch

packages/widget-v2/src/state/swapExecutionPage.ts
```diff
@@ -38,9 +40,10 @@ export const chainAddressesAtom = atom<Record<number, ChainAddress>>({});
 export const swapExecutionStateAtom = atomWithStorage<SwapExecutionState>(
   "swapExecutionState",
   {
-    route: undefined,
+    route: {} as RouteResponse,
```
codingki, line 43: I think it's ok to make it undefined, any reason to do this?

toddkao: yeah i forget why i did this tbh, I think having it be undefined makes sense

codingki (approved):

PR #270, opened 2024-09-24
feat: set manual address

codingki:


packages/widget-v2/src/pages/SwapPage/useSetMaxAmount.ts
codingki, line 1: gas substraction is only the source asset is the fee asset. fixed here

packages/widget-v2/src/hooks/useCreateEvmWallets.tsx
```diff
@@ -17,6 +16,9 @@ export const useCreateEvmWallets = () => {
   const currentConnector = connectors.find((connector) => connector.id === currentEvmConnector?.id);
   const createEvmWallets = useCallback((chainID: string) => {
     const isSei = chainID === "pacific-1";
+    if (isSei) {
+      chainID = sei.id.toString();
```
codingki, line 20: use evm chain id

packages/widget-v2/src/pages/SwapPage/SwapPage.tsx
```diff
@@ -296,46 +302,48 @@ export const SwapPage = () => {
...
+                      </TransparentButton>
+                    </>
+                  )
+                }
+              </Row >
             )
           }
         />
-        <Column align="center">
+        < Column align="center" >
```
toddkao, line 346: nit: what is this linter? lol why is it adding spaces like that

codingki: xd

PR #281, opened 2024-09-26
Misc updates 

toddkao:
Add storing atoms in localStorage
Add renderingSignaturesStillRequired
Remove using useSyncPendingTransactionHistoryItems
Add truncating amounts and chains names in case of very long strings
Add title so you can see truncated values when hovering cursor over it
Add copy to clipboard for addresses 
Fix arrow for SetAddressModal

truncating asset numbers

very long chain name


codingki (approved):

packages/widget-v2/src/state/swapPage.ts
```diff
@@ -86,9 +93,9 @@ export const invertSwapAtom = atom(null, (get, set) => {
   }
 });
 
-export const connectedWalletAtom = atom<MinimalWallet>();
+export const connectedWalletAtom = atomWithStorage<MinimalWallet | undefined>("connectedWallet", undefined);
 
-export const destinationWalletAtom = atom<MinimalWallet>();
+export const destinationWalletAtom = atomWithStorage<MinimalWallet | undefined>("destinationWallet", undefined);
```
codingki, line 98: I don't think we need to save this to local storage

PR #283, opened 2024-09-26
Pull all the balances

toddkao:
lift most valuable asset to top


packages/widget-v2/src/state/balances.ts
```diff
@@ -0,0 +1,23 @@
...
+  return {
+    queryKey: ["skipBalances", params],
+    queryFn: async () => {
+      if (!params) {
+        throw new Error("No balance request provided");
+      }
+
+      return skip.balances(params);
+    },
+    retry: 1,
```
codingki, line 21: can we add refetchInterval, maybe 30s or 1min or what ever you think it make sense

packages/widget-v2/src/pages/SwapPage/SwapPage.tsx
```diff
@@ -68,30 +68,30 @@ export const SwapPage = () => {
...
+          acc[asset.chainID].denoms.push(asset.denom);
+        } else {
+          acc[asset.chainID] = {
+            address: address,
+            denoms: [asset.denom],
+          };
+        }
+      }
+      return acc;
+    }, {} as Record<string, { address: string, denoms: string[] }>);
```
codingki, line 89: I think we can have this effect in `useSourceAccount`

codingki (approved):

packages/widget-v2/src/state/balances.ts
toddkao: I'll add 1 min for now, I'm a little worried about this endpoint xd

toddkao: maybe we should consider only refetching balances for a specific asset after you actually do a swap

PR #284, opened 2024-09-26
Widgetv2 bugfix

toddkao:
Bugfix for bug caused by fetching initial value of debounced amounts from localStorage,
instead add a hook that initializes debounced values


packages/widget-v2/src/hooks/useInitializeDebouncedValues.ts
codingki, line 1: run prettier

codingki (approved):
lgtm

PR #285, opened 2024-09-27
feat(widget-v2): grouping assets

codingki:


packages/widget-v2/src/modals/TokenAndChainSelectorModal/TokenAndChainSelectorModal.tsx
```diff
@@ -2,43 +2,159 @@ import { createModal, ModalProps, useModal } from "@/components/Modal";
...
     const [searchQuery, setSearchQuery] = useState<string>("");
 
-    const filteredAssets = useMemo(() => {
+    const [groupedAssetSelected, setGroupedAssetSelected] =
+      useState<GroupedAsset | null>(null);
+
+    const resetInput = () => {
+      setSearchTerm("");
+      setSearchQuery("");
+    };
```
toddkao, line 64: @codingki do we need both searchTerm and searchQuery?

packages/widget-v2/src/pages/SwapPage/SwapPage.tsx
```diff
@@ -111,27 +93,33 @@ export const SwapPage = () => {
         }));
         tokenAndChainSelectorModal.hide();
       },
+      selectedAsset: getClientAsset(sourceAsset?.denom, sourceAsset?.chainID),
```
toddkao, line 96: we don't want to pass selectedAsset for the asset selectors

packages/widget-v2/src/pages/SwapPage/SwapPage.tsx
```diff
@@ -144,26 +132,38 @@ export const SwapPage = () => {
         }));
         tokenAndChainSelectorModal.hide();
       },
+      selectedAsset: getClientAsset(
+        destinationAsset?.denom,
+        destinationAsset?.chainID
+      ),
```
toddkao, line 138: we don't want to pass selectedAsset for this one either (destinationAsset)

packages/widget-v2/src/pages/SwapPage/SwapPage.tsx
toddkao: ATOM
the selectedAsset is for rendering the asset in the search input but we only want that for the chain selector / chain selection step

toddkao (approved):
looks good besides the few comments i left

PR #287, opened 2024-09-27
Add fetching balances to main button

toddkao:
add fetching balances to main button


codingki:
I'm not ok with this, this make a duplicated balance loading state

PR #288, opened 2024-09-27
Fix fonts and remove stackledModal, add min-height for txhistoryModal

toddkao:


packages/widget-v2/src/components/Modal.tsx
```diff
@@ -86,7 +86,6 @@ export const useModal = <T extends ModalProps>(
       ...modalInstance,
       show: (showArgs?: Partial<T & ModalProps>) => {
         modalInstance.show({
-          stackedModal: numberOfModalsOpen > 0,
```
toddkao, line 89: just gonna remove this for now, it's meant purely for preventing double modals but for now we'll only render the widget directly so this is unnecessary

codingki (approved):

PR #291, opened 2024-09-27
release-v2-alpha

toddkao:


examples/nextjs/.eslintrc.json
codingki, line 1: is it safe to add back?

codingki (approved):

PR #297, opened 2024-10-01
Fix route taking forever on initial page load

toddkao:


packages/widget-v2/src/state/swapPage.ts
```diff
@@ -9,19 +9,36 @@ export type AssetAtom = Partial<ClientAsset> & {
...
+  (get) => {
+    return (
+      get(_debouncedDestinationAssetAmountAtom) ??
+      get(destinationAssetAtom)?.amount
+    );
+  },
+  (_get, set, newAmount: string) => {
+    set(_debouncedDestinationAssetAmountAtom, newAmount);
+  }
+);
```
toddkao, line 41: basically just fallback to returning the actual destinationAssetAtom?.amount if values are undefined

codingki:
eslint

PR #298, opened 2024-10-01
Refactor balances

toddkao:


packages/widget-v2/src/hooks/useFetchBalances.ts
```diff
@@ -1,20 +1,22 @@
...
   const { data: assets } = useAtomValue(skipAssetsAtom);
-  const setSkipBalancesRequest = useSetAtom(skipBalancesRequestAtom);
+  const setSkipAllBalancesRequest = useSetAtom(skipAllBalancesRequestAtom);
+  const skipBalances = useAtomValue(skipBalancesAtom);
 
   useEffect(() => {
     const allBalancesRequest = assets?.reduce((acc, asset) => {
       const address = getAccount(asset.chainID)?.address;
       if (address) {
+        if (asset.denom === sourceAsset?.denom) return acc;
```
toddkao, line 19: we're omitting the sourceAsset denom, since we're assuming it's being fetched by `useFetchSourceBalance`, honestly it's kinda unnecessary to do this it only prevents one extra denom balance from being fetched again

codingki (approved):

PR #302, opened 2024-10-01
Update error handling for route

toddkao:


packages/widget-v2/src/state/route.ts
```diff
@@ -0,0 +1,128 @@
...
+          allowMultiTx: true,
+          allowUnsafe: true,
+        });
+        return response;
+      } catch (error) {
+        return {
+          isError: true,
+          error,
+        };
+      }
```
toddkao, line 102: I'm basically wrapping this `Promise` with a try catch so i can return the error, Ideally i'd let atomWithQuery handle this but apparently it doesn't handle this :(

packages/widget-v2/src/state/route.ts
```diff
@@ -0,0 +1,128 @@
...
+      isLoading: false,
+    };
+  }
+  return {
+    data: routeResponse,
+    isError,
+    error,
+    isLoading,
+  };
+});
```
toddkao, line 128: we use this `skipRouteAtom` instead of the above to actually get the combined values and still maintain the same shape of the data when consuming the data, while getting the new error data now

codingki (approved):

PR #306, opened 2024-10-02
[FRE-1011, FRE-1012] feat: onTransactionSigned and onValidateGasBalance

codingki:
https://github.com/user-attachments/assets/a4139923-57de-4760-b1c1-de0915dc5747

### DIFF
--- a/packages/widget-v2/src/state/swapExecutionPage.ts
+++ b/packages/widget-v2/src/state/swapExecutionPage.ts
@@ -1,7 +1,7 @@
 import { atomWithMutation, atomWithQuery } from "jotai-tanstack-query";
 import { skipClient, skipRouteAtom } from "./skipClient";
 import { atom } from "jotai";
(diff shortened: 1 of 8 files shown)


packages/widget-v2/src/state/swapExecutionPage.ts
```diff
@@ -1,21 +1,30 @@
...
 import { atomEffect } from "jotai-effect";
 import { atomWithStorage } from "jotai/utils";
 import { setTransactionHistoryAtom, transactionHistoryAtom } from "./history";
 import { SimpleStatus } from "@/utils/clientType";
 import { errorAtom, ErrorType } from "./errorPage";
 
+type ValidatingGasBalanceData = {
+  chainID?: string;
+  txIndex?: number;
+  status: "success" | "error" | "pending" | "completed"
```
toddkao, line 15: every time i see us adding ID it hurts me, but we'll just have to update this later once we refactor the client library

toddkao (approved):

PR #316, opened 2024-10-03
add atomWithStorageNoCrossTabSync function

toddkao:
jotai's atomWithStorage by default will be subscribed to storage changes, which means cross-tab sync'd data
https://jotai.org/docs/utilities/storage#parameters
image

so in order to prevent this, we need to pass in a custom storage implementation:

image

there are some places where we can keep the cross-tab sync'd data turned on, so i just replaced the problematic ones


codingki (approved):
lgtm

PR #317, opened 2024-10-03
return initial value if no value found in localStorage

toddkao:
return initial value if no value found in localStorage

get rid of "--" if no usdValue is found


packages/widget-v2/src/pages/SwapPage/SwapPageAssetChainInput.tsx
```diff
@@ -188,7 +188,7 @@ export const SwapPageAssetChainInput = ({
               onMouseEnter={() => setShowPriceChangePercentage(true)}
               onMouseLeave={() => setShowPriceChangePercentage(false)}
             >
-              {(usdValue && formatUSD(usdValue)) ?? "--"}
+              {(usdValue && formatUSD(usdValue))}
```
toddkao, line 191: also let's remove the "--" it's uglier than rendering nothing

NotJeremyLiu (approved):

PR #319, opened 2024-10-03
Fix finding swapTransferEvent

toddkao:


packages/widget-v2/src/utils/clientType.ts
```diff
@@ -244,9 +244,15 @@ export function getOperationToTransferEventsMap(
         transferEvent.fromChainID === operation.fromChainID && transferEvent.toChainID === operation.toChainID
     );
 
-    const foundTransferEventWithTheSameFromChainId = transferEvents?.find(
-      (transferEvent) =>
-        transferEvent.fromChainID === operation.fromChainID
+    const foundSwapTransferEvent = transferEvents?.find(
```
ericHgorski, line 247: I maybe would go with something more like this for readability: 
```const foundSwapTransferEvent = transferEvents?.find((transferEvent) => {
  const isSwapType = ["evmSwap", "swap"].includes(operation.type);
  if (!isSwapType) return false;

  const sameChain = operation.fromChainID === operation.toChainID;
  const fromChainMatches = transferEvent.fromChainID === operation.fromChainID;
  const toChainMatches = transferEvent.toChainID === operation.toChainID;

  return sameChain
    ? fromChainMatches || toChainMatches
    : fromChainMatches && toChainMatches;
});
```

ericHgorski (approved):
Nit on confusing code but otherwise LGTM

packages/widget-v2/src/utils/clientType.ts
toddkao: I'll refactor it for readability

PR #320, opened 2024-10-04
[FRE-1052] Fix wallet name error when connecting wallet 

toddkao:


packages/widget-v2/src/hooks/useCreateCosmosWallets.tsx
```diff
@@ -90,12 +90,13 @@ export const useCreateCosmosWallets = () => {
         return address;
       };
       const walletInfo = getCosmosWalletInfo(wallet);
```
toddkao, line 92: `getCosmosWalletInfo` is from graz, ideally we should update the response type to be " | undefined" since sometimes it can return undefined

codingki (approved):

PR #321, opened 2024-10-04
FRE-1063, FRE-1103, FRE-1105

toddkao:
Screenshot 2024-10-04 151456


packages/widget-v2/src/pages/SwapExecutionPage/SwapExecutionPageRouteSimpleRow.tsx
```diff
@@ -49,13 +47,6 @@ export const SwapExecutionPageRouteSimpleRow = ({
     tokenAmount,
   });
 
-  const txStateOfAnimatedBorder = useMemo(() => {
-    if (destination && status === "broadcasted") {
-      return;
-    }
-    return status;
-  }, [status, destination]);
-
```
toddkao, line 58: broadcasted would not show any loading animation in animated border, this code is unncessary/redundant

packages/widget-v2/src/pages/SwapExecutionPage/useFetchTransactionStatus.tsx
```diff
@@ -24,86 +36,97 @@ export const useFetchTransactionStatus = () => {
...
+              toChainID: operation.toChainID,
+              fromChainID: operation.fromChainID,
+            } as ClientTransferEvent;
+
+            return accumulator;
+          },
+          {} as Record<number, ClientTransferEvent>
+        );
+        setOperationToTransferEventsMap(derivedOperationToTransferEventsMap);
+      }
```
toddkao, line 95: handling the case of not getting transfer event back but getting an overall status

packages/widget-v2/src/pages/SwapExecutionPage/useFetchTransactionStatus.tsx
```diff
@@ -24,86 +36,97 @@ export const useFetchTransactionStatus = () => {
...
-      clientOperations
-    );
-    const operationTransferEventsArray = Object.values(
-      operationToTransferEventsMap
-    );
 
     if (transactionDetailsArray.length === 1 && transferEvents.length === 0) {
+      // temporarily setting operationToTransferEventsMap to have the first
+      // operation pending before the API returns any transfer events
+      // so that the UX is nicer
```
toddkao, line 63: add comments to explain weird code

packages/widget-v2/src/state/swapExecutionPage.ts
```diff
@@ -21,7 +21,7 @@ type SwapExecutionState = {
   route?: RouteResponse;
   transactionDetailsArray: TransactionDetails[];
   transactionHistoryIndex: number;
-  overallStatus?: SimpleStatus;
+  overallStatus: SimpleStatus;
```
toddkao, line 24: I changed this so it can't be undefined, but instead returns "" empty string if it would other wise be undefined, it's kinda weird, but I mainly want to be able to use this:

```ts

["completed", "failed"].includes(overallStatus);
```
but we get type errors if overallStatus can be undefined

packages/widget-v2/src/utils/clientType.ts
```diff
@@ -338,11 +338,11 @@ export enum TransferType {
 }
 
 export type SimpleStatus =
+  | ""
+  | "signing"
```
toddkao, line 342: "broadcasted" was the old equivalent of "" since we weren't actually ever rendering anything for "broadcasted"

toddkao: there's an argument to be made that we should change "" to something else like "unconfirmed"

toddkao: actually i like that, ill update my code

packages/widget-v2/src/pages/SwapExecutionPage/SwapExecutionPageRouteDetailedRow.tsx
```diff
@@ -226,7 +226,6 @@ const StyledLoadingContainer = styled(Row) <{
         return `border: ${borderSize}px solid ${theme.success.text}`;
       case "failed":
         return `border: ${borderSize}px solid ${theme.error.text}`;
-      case "broadcasted":
```
toddkao, line 229: redundant, falls back to default case in every other case no need to be explicit

packages/widget-v2/src/pages/SwapExecutionPage/useFetchTransactionStatus.tsx
toddkao: it won't render an explorerLink for now but @NotJeremyLiu says this is fine for the time being and that V1 also doesn't render links for this case?

toddkao: also state toChainID and fromChainID honestly aren't really used at all typescript just complaining if i dont provide it

codingki:
Tested, swaps completed but still loading
image
Screenshot 2024-10-06 at 19 39 26

codingki (approved):

PR #323, opened 2024-10-04
[FRE-1021 FRE-1054] feat: connect eco

codingki:
https://github.com/user-attachments/assets/17849e2d-01b1-4fd4-bb99-3e8154e1a533


toddkao (approved):
nit: 
- need to fix the eslint / missing deps
- right arrow in the eco selector should have a white background so it's visible

otherwise lgtm

PR #328, opened 2024-10-07
[FRE-1068, FRE-1069] add modal animation

codingki:
https://github.com/user-attachments/assets/665805b5-a297-4fff-932e-8fed4065f081


toddkao:
looks good but I'm not sure if we should have these for all modals or only for drawer, I'm gonna ask jeremy and sonla

toddkao (approved):

PR #336, opened 2024-10-08
[FRE-1010, FRE-1074] Sort filteredChains by originChainID after balances

toddkao:
- Sort filtered chains by originChainID after balances
- Refactor balances fetching to only fetch all balances in one request, instead of fetching source balance sperately

image


packages/widget-v2/src/modals/TokenAndChainSelectorModal/TokenAndChainSelectorModal.tsx
```diff
@@ -207,6 +207,14 @@ export const TokenAndChainSelectorModal = createModal(
...
         }
 
+        if (assetB.asset.originChainID === assetB.chainID) {
+          return 1;
+        }
+
+        if (assetA.asset.originChainID === assetA.chainID) {
+          return -1;
+        }
+
```
toddkao, line 217: This is the sorting change, everything else is for consolidating the fetch balance back to one request, ever since the refactor I believe there's no longer much of a benefit to fetching balances seperately

codingki (approved):

codingki:
FYI this only applies to cosmos chain, because evm originChainId is not reliable

PR #337, opened 2024-10-08
[FRE-1017, FRE-1041] Add removing selected groupedAsset if using backspace when searchQuery is "", Add down arrow focusing VirtualList

toddkao:


codingki (approved):
can we add enter "down" key starts selecting the item, no need to enter "tab" key

packages/widget-v2/src/components/VirtualList.tsx
```diff
@@ -27,22 +27,25 @@ export const VirtualList = <T,>({
...
+    const firstElementInList = (listElement?.firstChild as HTMLElement);
 
     const handleKeyDown = (event: KeyboardEvent) => {
+
       if (event.key === "ArrowDown") {
+        if (firstElementInList) {
+          firstElementInList.focus();
+          setCurrentlyFocusedElement(firstElementInList);
+        }
+
```
toddkao, line 40: @codingki yeah this makes it select the first item in the list when pressing the down arrow

PR #347, opened 2024-10-09
[FRE-2049] Pass arbitrary cosmos msgs to be included before and after route execution

ericHgorski:


yarn.lock
codingki, line 1: can you discard this change and then yarn install again

codingki (changes requested):

yarn.lock
ericHgorski: thats what I tried to do, not sure why its modifying it at all, I didnt change any deps

codingki (approved):

PR #356, opened 2024-10-10
Simplify search on groupedAssets

toddkao:
Maybe we should just only search on the id / groupedSymbol? it will match expectations of what the search should do better
image

### DIFF
--- a/packages/widget-v2/src/modals/TokenAndChainSelectorModal/TokenAndChainSelectorModal.tsx
+++ b/packages/widget-v2/src/modals/TokenAndChainSelectorModal/TokenAndChainSelectorModal.tsx
@@ -157,12 +157,7 @@ export const TokenAndChainSelectorModal = createModal(
     const filteredAssets = useMemo(() => {
       if (!groupedAssetsByRecommendedSymbol) return;
       return matchSorter(groupedAssetsByRecommendedSymbol, searchQuery, {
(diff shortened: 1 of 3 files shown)


packages/widget-v2/src/modals/TokenAndChainSelectorModal/TokenAndChainSelectorModal.tsx
```diff
@@ -157,12 +157,7 @@ export const TokenAndChainSelectorModal = createModal(
     const filteredAssets = useMemo(() => {
       if (!groupedAssetsByRecommendedSymbol) return;
       return matchSorter(groupedAssetsByRecommendedSymbol, searchQuery, {
-        keys: [
-          "id",
-          "assets.*.symbol",
-          "assets.*.denom",
```
codingki, line 163: we can keep the denom

codingki (approved):

packages/widget-v2/src/modals/TokenAndChainSelectorModal/TokenAndChainSelectorModal.tsx
toddkao: @codingki the denom seems to match groups that the user probabbly wouldn't expect, for example with osm (the example in linear)

image

PR #358, opened 2024-10-10
[FRE-1018] speed up time to wallet signing

ericHgorski:
1. implement caching middleware to cache assets and chains calls (configurable with cacheDurationMs and cached based on request options)
2. reduce number of StargateClient instantiations
3. lightly refactor executeTxs to break out chains specific logic into their own functions. 
4. force optimize deps in vite to rebuild skip-go/client to improve testing flow

### DIFF
--- a/packages/client/src/__test__/client.test.ts
+++ b/packages/client/src/__test__/client.test.ts
@@ -12,8 +12,170 @@ import {
   RouteResponse,
   RouteResponseJSON,
 } from '../types';
+import { createHash } from 'crypto';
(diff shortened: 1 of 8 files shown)


ericHgorski:
Time to sign popup: 
before: 8328.5390625 ms
after: 6050.569091796875 ms

And some additional light touch code organization

codingki (approved):

codingki (changes requested):
I think we should add params to the cache key

codingki (approved):

PR #359, opened 2024-10-11
[FRE-1088, FRE-1008, FRE-1113, FRE-1110] fix history

codingki:


toddkao (approved):
nice! we can also delete the `useSyncPendingTransactionHistoryItems.tsx` file too

PR #363, opened 2024-10-11
Add rendering Timeout error page if time takes > 3 times estimated time

toddkao:
Testing with 5 second timeout (to force it to show)

https://github.com/user-attachments/assets/34d909f7-9afc-4d8e-96e6-0cb05204ec5c


packages/widget-v2/src/pages/SwapExecutionPage/useHandleTransactionTimeout.tsx
```diff
@@ -0,0 +1,36 @@
...
+  const { route, transactionDetailsArray, } = useAtomValue(
+    swapExecutionStateAtom
+  );
+  const setError = useSetAtom(errorAtom);
+  const setOverallStatus = useSetAtom(setOverallStatusAtom);
+  const [transactionTimeoutTimer, setTransactionTimeoutTimer] = useState<NodeJS.Timeout | undefined>();
+
+  useEffect(() => {
+    if (!route?.estimatedRouteDurationSeconds) return;
+    if (swapExecutionState === SwapExecutionState.pending && transactionTimeoutTimer === undefined) {
```
codingki, line 17: add `&& route.txRequired === transactionDetailsArray.length`

toddkao: hmm

toddkao: if we want to change the logic so we wait until we get all of the transactionDetails before we start the timer, I think we can reduce the time to be like 1x estimatedRouteDuration

codingki: yeah I agree, because it's better to make sure user have signed all tx

packages/widget-v2/src/pages/SwapExecutionPage/useHandleTransactionTimeout.tsx
```diff
@@ -0,0 +1,36 @@
...
+          explorerLink: lastTransaction?.explorerLink ?? "",
+        });
+      }, route.estimatedRouteDurationSeconds * 1_000 * 3);
+
+      setTransactionTimeoutTimer(timeoutTimer);
+    }
+    if (swapExecutionState === SwapExecutionState.confirmed && transactionTimeoutTimer !== undefined) {
+      clearTimeout(transactionTimeoutTimer);
+    }
+  }, [route, setError, setOverallStatus, swapExecutionState, transactionDetailsArray, transactionTimeoutTimer]);
```
codingki: please add cleanup fn
```
return () => {
      clearTimeout(timeoutTimer);
    };

```

packages/widget-v2/src/pages/SwapExecutionPage/useHandleTransactionTimeout.tsx
toddkao: yeah that makes sense

packages/widget-v2/src/pages/SwapExecutionPage/useHandleTransactionTimeout.tsx
toddkao: good catch

codingki (approved):

PR #374, opened 2024-10-23
Update SwapPage to use calculatePercentageChange from @/utils/number to be consistent

toddkao:
Updated `calculatePercentageDifference` function name to be `calculatePercentageChange`
Updated `SwapPage` to use `calculatePercentageChange`

image

image

### DIFF
--- a/packages/widget-v2/src/utils/number.ts
+++ b/packages/widget-v2/src/utils/number.ts
@@ -34,7 +34,7 @@ export function limitDecimalsDisplayed(input: string | number) {
   return flooredAndLimitedDecimalPlacesNumber.toString();
 }
 
-export function calculatePercentageDifference(
+export function calculatePercentageChange(
   numberA: number | string,
   numberB: number | string,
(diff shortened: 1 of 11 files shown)


packages/widget-v2/src/utils/number.ts
```diff
@@ -34,7 +34,7 @@ export function limitDecimalsDisplayed(input: string | number) {
   return flooredAndLimitedDecimalPlacesNumber.toString();
 }
 
-export function calculatePercentageDifference(
+export function calculatePercentageChange(
```
toddkao, line 37: This function calculates the percentage change

$$
\frac{(V_2 - V_1)}{|V_1|} \times 100
$$

not the percentage difference

$$
\frac{|V_1 - V_2|}{\left(\frac{V_1 + V_2}{2}\right)} \times 100
$$

which is what caused the confusion/difference in calculation

codingki (approved):
can we add `text-wrap: balance;` to the error message as well in this PR

PR #380, opened 2024-10-25
Add client side balance checking

codingki:


codingki:
Cancelled because we fixed balance endpoint

PR #384, opened 2024-10-28
[FRE-1086, FRE-1118, FRE-1155]

ericHgorski:
1. vertically center and properly size estimated time on swap
2. swap complete -> swap again button text
3. use correct solana logo
4. only show background for buttons on hover


toddkao (approved):
lgtm

PR #389, opened 2024-10-28
Fix duplicate chains

toddkao:
image


packages/widget-v2/src/modals/TokenAndChainSelectorModal/TokenAndChainSelectorModal.tsx
```diff
@@ -187,15 +187,22 @@ export const TokenAndChainSelectorModal = createModal(
...
         })
-        .filter((c) => c) as ChainWithAsset[];
+        .filter((chain) => {
+          if (!chain) return false;
+          if (seenChainIds.has(chain.chainID)) return false;
+
+          seenChainIds.add(chain.chainID);
+          return true;
+        }) as ChainWithAsset[];
+
```
toddkao, line 205: I need to think about how to handle different assets with the same recommended symbol and chain ID

toddkao: this is naive and just takes whichever one with the same chainID is first in the aray

PR #393, opened 2024-10-29
Update debounced value initialization implementation

toddkao:
use `initializeDebounceValuesEffect` to handle initializing debounced values instead of previous method
also, fix slippage stuff


packages/widget-v2/src/pages/SwapPage/SwapPage.tsx
```diff
@@ -67,12 +68,15 @@ export const SwapPage = () => {
 
   const setChainAddresses = useSetAtom(chainAddressesAtom);
   useFetchAllBalances();
+  useCleanupDebouncedAtoms();
```
toddkao, line 71: only change to this file

ericHgorski (approved):
LGTM

PR #394, opened 2024-10-30
Fix stack size exceeded error and add missing dep (for yarn install)

toddkao:


packages/widget-v2/package.json
```diff
@@ -23,8 +23,8 @@
     "README.md"
   ],
   "peerDependencies": {
-    "react": "17.x || 18.x",
-    "react-dom": "17.x || 18.x"
+    "react": ">=17.0.0",
+    "react-dom": ">=17.0.0"
```
toddkao, line 27: this should fix installation errors with react 19

packages/widget-v2/package.json
```diff
@@ -60,6 +60,7 @@
     "@ebay/nice-modal-react": "^1.2.13",
     "@eslint/compat": "^1.1.1",
     "@initia/initia-registry": "^0.1.16",
+    "@leapwallet/cosmos-social-login-capsule-provider": "^0.0.44",
```
toddkao, line 63: this fixes yarn install, it's the only extra dep that you'd need to install manually when using yarn

packages/widget-v2/package.json
```diff
@@ -79,7 +80,7 @@
     "@tanstack/react-query": "^5.51.21",
     "chain-registry": "^1.63.43",
     "graz": "^0.1.25",
-    "jotai": "^2.9.1",
+    "jotai": "^2.10.1",
```
toddkao, line 83: updated jotai version

packages/widget-v2/src/state/swapPage.ts
```diff
@@ -128,38 +125,6 @@ export const invertSwapAtom = atom(null, (get, set) => {
...
-      amount: limitDecimalsDisplayed(swapInAmount),
-    }));
-  } else if (direction === "swap-out" && swapOutAmountChanged) {
-    set(sourceAssetAtom, (old) => ({
-      ...old,
-      amount: limitDecimalsDisplayed(swapOutAmount),
-    }));
-  }
-});
-
```
toddkao, line 162: I believe this was the main culprit causing the stack size exceeded / infinite set state

packages/widget-v2/package.json
toddkao: interestingly, after updating the jotai version i was able to see the stack size exceeded in the devmode / http://localhost:5173

packages/widget-v2/src/state/swapPage.ts
codingki: I see, nice catch

codingki (approved):

PR #402, opened 2024-10-31
[FRE-1161] remove debounce to simplify code and clear inputs on route complete or reload

ericHgorski:


toddkao (changes requested):
I stand by the fact that we shouldn't get rid of the debouncing, I'm fine with re-implementing in another way but not removing it

packages/widget-v2/src/widget/Widget.tsx
```diff
@@ -89,16 +93,28 @@ const WidgetWithoutNiceModalProvider = (props: WidgetProps) => {
     setTheme(mergedTheme);
   }, [setSkipClientConfig, mergedSkipClientConfig, setTheme, mergedTheme]);
 
+  useLayoutEffect(() => {
```
ericHgorski, line 96: @toddkao

PR #403, opened 2024-10-31
Fix 0 balance, fix undefined symbol

toddkao:
- Fix 0 balance
- Fix undefined symbol
- Disable autoreconnect in GrazProvider

### DIFF
--- a/packages/widget-v2/src/hooks/useGetSourceBalance.ts
+++ b/packages/widget-v2/src/hooks/useGetSourceBalance.ts
@@ -19,7 +19,11 @@ export const useGetSourceBalance = () => {
     const sourceBalance = skipBalances?.chains?.[chainID]?.denoms?.[denom];
 
     if (denomsExists && sourceBalance === undefined) {
-      return null;
+      return {
+        amount: 0,
+        formattedAmount: "0",
+        error: undefined,
+      };
     }
     return skipBalances?.chains?.[chainID]?.denoms?.[denom];
   }, [sourceAsset, sourceAccount, skipBalances]);
--- a/packages/widget-v2/src/hooks/useGetSourceBalance.ts
+++ b/packages/widget-v2/src/hooks/useGetSourceBalance.ts
@@ -15,6 +15,12 @@ export const useGetSourceBalance = () => {
     const { chainID, denom } = sourceAsset;
     if (!denom || !chainID) return;
 
+    const denomsExists = !!skipBalances?.chains?.[chainID]?.denoms;
+    const sourceBalance = skipBalances?.chains?.[chainID]?.denoms?.[denom];
+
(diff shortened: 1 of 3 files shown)


packages/widget-v2/src/pages/SwapPage/ConnectedWalletContent.tsx
```diff
@@ -28,8 +28,10 @@ export const ConnectedWalletContent = () => {
   const connectedWalletModal = useModal(ConnectedWalletModal);
 
   const formattedBalance = useMemo(() => {
+    const symbol = sourceDetails?.symbol ? sourceDetails?.symbol : "";
     if (sourceBalance?.error?.message) return "--";
-    if (sourceBalance === undefined) return `0 ${sourceDetails?.symbol}`;
+    if (sourceBalance === undefined) return;
+    if (sourceBalance === null) return `0 ${symbol}`;
```
toddkao, line 34: null and undefined are different, null means 0 balance, undefined means balance is not fetched yet

toddkao: this is because when we don't pass in any denoms, the server will not return back the denoms for the balances that are 0

packages/widget-v2/src/hooks/useGetSourceBalance.ts
```diff
@@ -15,6 +15,12 @@ export const useGetSourceBalance = () => {
     const { chainID, denom } = sourceAsset;
     if (!denom || !chainID) return;
 
+    const denomsExists = !!skipBalances?.chains?.[chainID]?.denoms;
+    const sourceBalance = skipBalances?.chains?.[chainID]?.denoms?.[denom];
+
+    if (denomsExists && sourceBalance === undefined) {
+      return null;
+    }
```
toddkao, line 23: this is because when we don't pass in any denoms, the server will not return back the denoms for the balances that are 0

packages/widget-v2/src/hooks/useGetSourceBalance.ts
```diff
@@ -15,6 +15,16 @@ export const useGetSourceBalance = () => {
...
+    const denomsExists = !!skipBalances?.chains?.[chainID]?.denoms;
+    const sourceBalance = skipBalances?.chains?.[chainID]?.denoms?.[denom];
+
+    if (denomsExists && sourceBalance === undefined) {
+      return {
+        amount: 0,
+        formattedAmount: "0",
+        error: undefined,
+      };
+    }
```
toddkao, line 27: this is because when we don't pass in any denoms, the server will not return back the denoms for the balances that are 0, but if `skipBalances?.chains?.[chainID]?.denoms` is found we can assume the server checked those denoms, but determined they were 0

codingki (approved):

PR #404, opened 2024-10-31
[FRE-1035] widget v2 migration guide and updates configs

ericHgorski:


docs/widget/migration-guide.mdx
```diff
@@ -0,0 +1,146 @@
...
+    theme='{
+        "backgroundColor": "#191A1C",
+        "textColor": "#E6EAE9",
+        "borderColor": "#363B3F",
+        "brandColor": "#FF4FFF",
+        "highlightColor": "#1F2022"
+    }'
+  />
+```
+**After:**
```
toddkao, line 34: btw, there's a new "brandColor" prop on the widget I think we should consider removing brandColor from the theme prop examples (even though it's valid) so that it doesn't confuse people and push people towards adding brandColor outside of the theme prop

the reason I had to add this was because if we want you to be able to use theme="light" / theme="dark" prop while also passing a brandColor

ericHgorski: ok added in an line comment, good shout!

toddkao (approved):

NotJeremyLiu (approved):
so fire sir, full send

PR #407, opened 2024-11-01
[FRE-1179] fix: multiple tx pending causing stuck loading

codingki:


NotJeremyLiu:
Does this solve:
https://linear.app/skip/issue/FRE-1179/status-confirmation-bug

toddkao (approved):

PR #408, opened 2024-11-01
Update swapVenues chainID -> chainId

toddkao:
I honestly doubt this prop is going to be used much or even at all, but I want to be consistent on the naming

### DIFF
--- a/packages/widget-v2/src/widget/Widget.tsx
+++ b/packages/widget-v2/src/widget/Widget.tsx
@@ -14,7 +14,12 @@ import {
   defaultSkipClientConfig,
   onlyTestnetsAtom,
 } from "@/state/skipClient";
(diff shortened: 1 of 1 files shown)


codingki (approved):

ericHgorski:
> I honestly doubt this prop is going to be used much or even at all, but I want to be consistent on the naming

Whats the plan with the temp one? Its just for backwards compatability?

toddkao:
> > I honestly doubt this prop is going to be used much or even at all, but I want to be consistent on the naming
> 
> Whats the plan with the temp one? Its just for backwards compatability?

@ericHgorski yeah this is here temporarily before we change the syntax on the client package, we can remove this once we do that, I think we'll change the syntax on the client package when we get to the client refactor

PR #416, opened 2024-11-04
[FRE-1181] fix: custom swap settings ux behavior

ericHgorski:
also externalized fee type checking logic and refactor slippage logic as its own component

### DIFF
--- a/packages/widget-v2/src/utils/route.ts
+++ b/packages/widget-v2/src/utils/route.ts
@@ -1,0 +1,58 @@
+import { formatUSD } from "@/utils/intl";
+import { EstimatedFee, RouteResponse } from "@skip-go/client";
+
+export enum FeeType {
+  SMART_RELAY = "SMART_RELAY",
+}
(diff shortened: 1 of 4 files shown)


packages/widget-v2/src/utils/route.ts
```diff
@@ -0,0 +1,57 @@
...
+
+  if (!computedAmount || !computedUsd) return undefined;
+
+  const assetDecimals = relayFees[0].originAsset.decimals || 6;
+  const inAsset = (computedAmount / Math.pow(10, assetDecimals)).toLocaleString(
+    "en-US",
+    {
+      maximumFractionDigits: 6,
+    }
+  );
```
toddkao, line 50: we have util functions in utils/crypto.ts for this type of thing

`convertHumanReadableAmountToCryptoAmount` and `convertTokenAmountToHumanReadableAmount`

toddkao (approved):

PR #419, opened 2024-11-04
fix destination wallet issues

codingki:


toddkao (approved):
nit: wonder why the linter changed " to ' ?

PR #426, opened 2024-11-05
Modal refactor

toddkao:
Refactor Modals


toddkao:
one big thing I noticed is if we open two modals and the second modal is smaller than the first one, this looks very weird 
image

codingki (approved):

PR #434, opened 2024-11-07
Kiki/fre 1204 got this view on mintscan link for a solana tx

codingki:


toddkao:
@codingki do you mind hiding the link if explorerLink is invalid?

packages/widget/src/utils/explorerLink.ts
```diff
@@ -0,0 +1,26 @@
...
+      const url = chain?.blockExplorers?.default.url
+      return url ? `${url}/tx/${txHash}` : txHash
+    }
+    case "svm": {
+      return `https://solscan.io/tx/${txHash}`
+    }
+    default:
+      return txHash
+  }
+}
```
toddkao, line 26: @codingki it doesn't seem obvious to me that `txHash` can be returned from this function,
I think we should either remove `txHash` from here and pass it separately to `ErrorPageTimeout`, or rename this function to make it clear that the `txHash` is potentially returned, something like "getExplorerLinkOrTxHash"

toddkao (approved):

PR #436, opened 2024-11-08
[FRE-1212] Update chainIdsToAffiliates prop to be camelCase, update web-component publish script/version

toddkao:
Update chainIdsToAffiliates prop to be camelCase
Unrelated but also updated web-component publish script to remove alpha tag and bumped version to match widget


codingki (approved):
lgtm

PR #456, opened 2024-11-12
Fix light mode modals 

toddkao:
old:
Screenshot 2024-11-12 at 10 09 08 AM

new:
Screenshot 2024-11-12 at 10 16 40 AM

### DIFF
--- a/packages/widget/src/components/Modal.tsx
+++ b/packages/widget/src/components/Modal.tsx
@@ -11,8 +11,9 @@ import {
 import { PartialTheme } from "@/widget/theme";
 
 import { ErrorBoundary } from "react-error-boundary";
-import { useAtom } from "jotai";
+import { useAtomValue, useSetAtom } from "jotai";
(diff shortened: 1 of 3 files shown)


codingki (approved):

packages/widget/src/components/Modal.tsx
```diff
@@ -84,10 +85,11 @@ export const createModal = <T extends ModalProps>(
   const Component = component;
 
   const WrappedComponent = (props: T) => {
-    const [, setError] = useAtom(errorAtom);
+    const setError = useSetAtom(errorAtom);
+    const theme = useAtomValue(themeAtom);
 
     return (
-      <Modal {...props}>
+      <Modal {...props} theme={theme}>
```
toddkao, line 92: once we pass `theme` from `themeAtom` here we can use `useTheme()` from `styled-components` as usual (which is what all of the modals are currently doing)

PR #459, opened 2024-11-12
[FRE-1224, FRE-1225] add tx details on tracked and precent status call until tracking is complete

ericHgorski:


codingki:
changeset

codingki (approved):

PR #481, opened 2024-11-19
improve cosmos balance validation

codingki:


packages/widget/src/hooks/useCosmosFeeAssetValidation.ts
```diff
@@ -1,54 +1,74 @@
...
+        if (!a.gasPrice) return undefined;
+        let price = a.gasPrice.average;
+        if (price === '') {
+          price = a.gasPrice.high;
+        }
+        if (price === '') {
+          price = a.gasPrice.low;
+        }
+        return new GasPrice(Decimal.fromUserInput(price, 18), a.denom);
+      })();
```
ericHgorski, line 34:
```suggestion
      const gasPrice = (() => {
        if (!a.gasPrice) return undefined;
        const price =
          a.gasPrice.average || a.gasPrice.high || a.gasPrice.low;
        return new GasPrice(Decimal.fromUserInput(price, 18), a.denom);
      })();
```

ericHgorski: I think this is more clear

ericHgorski: do we need the IIFE? or could we simply further with: 

```
let price;
if (a.gasPrice) {
  price = a.gasPrice.average || a.gasPrice.high || a.gasPrice.low;
}

return price
  ? new GasPrice(Decimal.fromUserInput(price, 18), a.denom)
  : undefined;
  ```

PR #486, opened 2024-11-20
[FRE-1237] have more intelligent gas limit estimations per chain tx

codingki:


ericHgorski (approved):
LGTM! Thank you for changing those to useAtomValue that was bothering me too lol

PR #488, opened 2024-11-20
[FRE-1248] style updates if go fast

ericHgorski:


packages/widget/src/components/MainButton.tsx
```diff
@@ -14,6 +16,13 @@ export type MainButtonProps = {
   loading?: boolean;
   loadingTimeString?: string;
   onClick?: () => void;
+  extra?: ReactNode;
+  route?: RouteResponse
+  useBrandColorForLoadingAnimation?: boolean;
```
codingki, line 21: Because currently this will used only for go fast I suggest to change the prop name
```suggestion
  isGoFast?: boolean;
```

packages/widget/src/components/MainButton.tsx
```diff
@@ -155,26 +172,32 @@ const StyledMainButton = styled(Row).attrs({
     `};
 `;
 
-const StyledLoadingButton = styled(StyledMainButton)`
+const StyledLoadingButton = styled(StyledMainButton)
+  <{ useBrandColor?: boolean; }>`
```
codingki, line 176: `isGoFast`

codingki (approved):

PR #495, opened 2024-11-21
[FRE-1255] go fast fee and route preference setting

ericHgorski:


packages/widget/src/state/route.ts
```diff
@@ -87,6 +87,7 @@ export const routeConfigAtom = atom<WidgetRouteConfig>({
     splitRoutes: true,
     evmSwaps: true,
   },
+  goFast: true
```
ericHgorski, line 90: Is this ok to add here as a default or should I remove @codingki ?

codingki: it's ok

codingki (commented):
changeset!
LGTM, nice refactor tyty ser

codingki (approved):

PR #498, opened 2024-11-22
feat: add onTransactionHash callback and improve transaction handling

ericHgorski:


packages/client/src/client.ts
```diff
@@ -502,33 +497,38 @@ export class SkipClient {
...
     signer,
-    onTransactionSigned,
-    onApproveAllowance,
+    options,
   }: {
     message: types.EvmTx;
     signer: WalletClient;
-    onTransactionSigned?: types.TransactionCallbacks['onTransactionSigned'];
-    onApproveAllowance?: types.TransactionCallbacks['onApproveAllowance'];
+    options: clientTypes.ExecuteRouteOptions;
```
codingki, line 524:
```suggestion
    callbacks: types.TransactionCallbacks
```

packages/client/src/client.ts
```diff
@@ -599,29 +605,33 @@ export class SkipClient {
   async executeSVMTransaction({
     signer,
     message,
-    onTransactionSigned,
+    options,
   }: {
     signer: Adapter;
     message: types.SvmTx;
-    onTransactionSigned?: types.TransactionCallbacks['onTransactionSigned'];
+    options: clientTypes.ExecuteRouteOptions;
```
codingki, line 612:
```suggestion
    callbacks: types.TransactionCallbacks
```

packages/client/src/client.ts
```diff
@@ -502,33 +497,38 @@ export class SkipClient {
...
 
     const txHash = toHex(sha256(txBytes));
-    onTransactionSigned?.({
-      chainID,
+
+    onTransactionHash?.({
       txHash,
+      chainID,
     });
+
```
codingki, line 512: move it after `broadcastTX`

packages/client/src/client-types.ts
```diff
@@ -112,6 +112,7 @@ export type ExecuteCosmosMessage = GasOptions & {
   messages: types.CosmosMsg[];
   gas: Gas
   onTransactionSigned?: TransactionCallbacks['onTransactionSigned'];
+  onTransactionHash?: TransactionCallbacks['onTransactionHash'];
```
codingki, line 115: Let's rename this to `onTransactionSubmitted`

codingki (changes requested):
I think we can use `onTransactionBroadcast` for this. Let's remove additional callback and use `onTransactionBroadcast` to replace `onTransactionTracked`

ericHgorski:
Just pushed those previous suggestions @codingki , let me know if you have further guidance!

codingki:
Remove `onTransactionSubmitted` callback in executeRoute. we can use `onTransactionSigned` or `onTransactionBroadcast`. because `onTransactionSubmitted` is basically the same with `onTransactionBroadcast`

ericHgorski:
> Remove `onTransactionSubmitted` callback in executeRoute. we can use `onTransactionSigned` or `onTransactionBroadcast`. because `onTransactionSubmitted` is basically the same with `onTransactionBroadcast`

Resolved!

codingki:
I don't think you need anything for `onTransactionBroadcast` because it's already in client.ts line 315 @ericHgorski

ericHgorski:
> client.ts

ah ok I see your point, yea makes sense!

codingki (approved):
turns out the options we pass is the executeRoute options, I change it to options instead of callback and fix the type.

merging this, tyty

PR #503, opened 2024-11-26
refactor: modal registration after WalletProvider

codingki:
- Share WalletProvider instead have 2 WalletProvider instance between widget and modal
- Fix bad setState in useGetAccount


ericHgorski:
These code improvements look awesome @codingki 

Anything in particular that should be tested regarding them? Or is it mostly quality of life improvement?

PR #505, opened 2024-11-26
Keep previous balance data to avoid re-loading/blocking user interaction

toddkao:


codingki (approved):
run changeset before merge

PR #508, opened 2024-11-26
refactor input handling for decimal values and update input type to text

ericHgorski:


packages/widget/src/pages/SwapPage/SwapPageAssetChainInput.tsx
```diff
@@ -57,13 +57,25 @@ export const SwapPageAssetChainInput = ({
...
+    // Keep only the first dot, remove any additional dots
+    const firstDotIndex = latest.indexOf('.');
+    if (firstDotIndex !== -1) {
+      latest =
+        latest.substring(0, firstDotIndex + 1) +
+        latest.substring(firstDotIndex + 1).replace(/\./g, '');
+    }
+
+    // Remove leading zeros unless they are immediately followed by a dot
+    latest = latest.replace(/^0+(?!\.)/, '0');
```
toddkao, line 75: I wonder if the validation code in `handleInputChange` is still necessary if we use the pattern attribute, i imagine it's possible to also limit decimals displayed in regex too

PR #526, opened 2024-12-02
Fix no route when inverting swap before entering amount

toddkao:
Fix no route when inverting swap before entering amountIn by removing the if conditions on actually swapping the two amount values


packages/widget/src/state/swapPage.ts
```diff
@@ -100,19 +100,15 @@ export const invertSwapAtom = atom(null, (get, set) => {
...
-      set(swapDirectionAtom, newSwapDirection);
-      set(isInvertingSwapAtom, false);
-    });
-  }
+  set(destinationAssetAmountAtom, sourceAsset?.amount ?? "", () => {
+    const newSwapDirection =
+      swapDirection === "swap-in" ? "swap-out" : "swap-in";
+    set(swapDirectionAtom, newSwapDirection);
+    set(isInvertingSwapAtom, false);
+  });
```
toddkao, line 111: the main issue is that if `sourceAsset?.amount` was undefined before the entire callback would not happen

codingki (approved):
lgtm

PR #530, opened 2024-12-02
fix: update height prop in StyledNoResultsContainer for better layout

ericHgorski:


toddkao (commented):
looks good just missing changeset

PR #534, opened 2024-12-03
[FRE-1247, FRE-1293] fix: improve balance loading state handling and display

ericHgorski:
also remove confusing tildas from error ouput


codingki (approved):
lgtm

PR #540, opened 2024-12-04
fix: update React and ReactDOM to version 18.2.0 and add useQueryPara…

ericHgorski:
…ms hook for handling URL query parameters


codingki (approved):
lgtm

PR #544, opened 2024-12-05
Refactor chain type handling to use enum for ChainType across client and widget packages

ericHgorski:


packages/widget/src/modals/ConnectedWalletModal/ConnectedWalletModal.tsx
```diff
@@ -69,16 +70,9 @@ const ConnectEco = ({
     assetDenom: sourceAsset?.denom,
     chainId: sourceAsset?.chainID,
   });
+
   const account = useMemo(() => {
-    const _chainID = (() => {
```
ericHgorski, line 73: Tried to make improvement here @codingki , let me know if you think this is problematic

packages/widget/src/hooks/useAutoSetAddress.ts
```diff
@@ -60,11 +61,11 @@ export const useAutoSetAddress = () => {
         if (!chain) {
           return;
         }
+        if (chainAddresses[index]?.address) return;
```
ericHgorski, line 64: Made a small code improvement here. Its better to just handle this at top of hook instead of 3 times for each case? 
@codingki

ericHgorski:
I took another look and dont think this is breaking @codingki. I dont see us passing "svm" "evm" etc. to client or widget in their current state.

codingki (approved):
LGTM! just need chageset

PR #555, opened 2024-12-09
refactor: update chainType to use ChainType enum in routing and wallet hooks

ericHgorski:


packages/client/src/types/routing.ts
```diff
@@ -46,7 +46,7 @@ export type Chain = {
   logoURI?: string;
   bech32Prefix: string;
   feeAssets: FeeAsset[];
-  chainType: ChainType;
+  chainType: ChainType | string;
```
codingki, line 49: let's keep using enum
```suggestion
  chainType: ChainType;
```

packages/client/src/types/routing.ts
```diff
@@ -69,7 +69,7 @@ export type ChainJSON = {
   logo_uri?: string;
   bech32_prefix: string;
   fee_assets: FeeAssetJSON[];
-  chain_type: string;
+  chain_type: ChainType | string;
```
codingki, line 72: let's keep using enum
```suggestion
  chainType: ChainType;
```

PR #558, opened 2024-12-10
Add callbacks

toddkao:
Add callbacks

- onWalletConnected
- onWalletDisconnected
- onTransactionBroadcasted
- onTransactionComplete
- onTransactionFailed


packages/widget/src/hooks/useCreateCosmosWallets.tsx
```diff
@@ -29,12 +29,14 @@ import { useCallback } from "react";
 import { skipAssetsAtom, skipChainsAtom } from "@/state/skipClient";
 import { sourceAssetAtom } from "@/state/swapPage";
 import { isMobile } from "@/utils/os";
+import { allCallbacksAtom } from "@/state/callbacks";
```
ericHgorski, line 32: In terms of naming, what do you think about just `callbacks` or `walletCallbacks` ? (for this atom as well as rest of references)

packages/widget/src/hooks/useCreateCosmosWallets.tsx
```diff
@@ -101,6 +103,7 @@ export const useCreateCosmosWallets = () => {
           },
           disconnect: async () => {
             console.error("Prax wallet is not supported");
+            allCallbacks?.onWalletDisconnected?.({ chainType: "cosmos" });
```
ericHgorski, line 106: use `ChainType` enum from client lib

packages/widget/src/hooks/useCreateCosmosWallets.tsx
```diff
@@ -260,6 +286,7 @@ export const useCreateCosmosWallets = () => {
           disconnect: async () => {
             await disconnectAsync();
             setCosmosWallet(undefined);
+            allCallbacks?.onWalletDisconnected?.({ chainType: "cosmos" });
```
ericHgorski, line 289: use `ChainType` enum from client lib

packages/widget/src/hooks/useCreateSolanaWallets.tsx
```diff
@@ -71,14 +86,14 @@ export const useCreateSolanaWallets = () => {
         disconnect: async () => {
           await wallet.disconnect();
           setSvmWallet(undefined);
-          // TODO: onWalletDisconnected
+          allCallbacks?.onWalletDisconnected?.({ chainType: "svm" });
```
ericHgorski, line 89: use `ChainType` enum from client lib

packages/widget/src/hooks/useCreateCosmosWallets.tsx
codingki: agree on this one

packages/widget/src/state/callbacks.ts
```diff
@@ -0,0 +1,48 @@
...
+  txHash: string;
+  chainId: string;
+  explorerLink: string;
+};
+
+export type OnTransactionFailedProps = {
+  error: string;
+};
+
+export type AllCallbacks = {
```
codingki, line 36: rename to Callbacks

ericHgorski:
Im not sure we should pass chainType to on wallet disconnect, as this gets confusing with multi-chain wallets (e.g. Keplr currently gets chainType: evm)

could we pass out wallet name instead?

ericHgorski:
Also for on wallet connectedI think its fine if we just pass `chainIdToAddressMap` and not `chainIds` as they can derive `chainIds` easily from the map. what do you think?

ericHgorski:
All the transaction stuff worked great!

toddkao:
> Im not sure we should pass chainType to on wallet disconnect, as this gets confusing with multi-chain wallets (e.g. Keplr currently gets chainType: evm)
> 
> could we pass out wallet name of disconnected wallet instead?

I think passing wallet name when disconnecting makes sense, but I also think specifying chainType also makes sense, as you said, you could have Keplr connected for both cosmos and EVM, when you disconnect one of them I think the idea is that if we didn't return chainType you wouldn't know if you disconnected Keplr for cosmos chains or for EVM chains

ericHgorski:
> > Im not sure we should pass chainType to on wallet disconnect, as this gets confusing with multi-chain wallets (e.g. Keplr currently gets chainType: evm)
> > could we pass out wallet name of disconnected wallet instead?
> 
> I think passing wallet name when disconnecting makes sense, but I also think specifying chainType makes sense, as you said, you could have Keplr connected for both cosmos and EVM, when you disconnect one of them I think the idea is that if we didn't return chainType you wouldn't know if you disconnected Keplr for cosmos chains or for EVM chains

Sounds good im aligned

ericHgorski:
> > Also for on wallet connectedI think its fine if we just pass `chainIdToAddressMap` and not `chainIds` as they can derive `chainIds` easily from the map. what do you think?
> 
> I don't think it's a big deal to leave it as is, but I don't feel too strongly about it

Ok lets remove it then! If I saw both of those it would make me think they were derived differently and I would feel confused. 

The only last thing that would be a nice to have is destructuring the props in the next-js example so if I was developer reading it, I could easily tell what info args I get in the callback instead of `props` in general. Otherwise LGTM! thanks for this

ericHgorski (approved):
LGTM!

PR #561, opened 2024-12-11
QA: fix edit destination for 2 tx routes

ericHgorski:


packages/widget/src/hooks/useAutoSetAddress.ts
```diff
@@ -46,10 +45,12 @@ export const useAutoSetAddress = () => {
     route?.requiredChainAddresses?.forEach((chainID, index) => {
       res[index] = {
         chainID,
+        address: chainAddresses[index]?.address,
+        chainType: chainAddresses[index]?.chainType,
```
codingki, line 49: this is unnecessary because these value is always undefined this will ran at initial step so should be undefined. Did you find a case that these value is not undefined?

ericHgorski: yes, it overrides the previously set values if this isnt specified, because this hook is called again after the 2 signature warning

codingki: I see, nice catch

codingki (approved):

PR #578, opened 2024-12-16
[FRE-1330] Don't auto focus / pop up keyboard on asset selection on mobile

toddkao:


packages/widget/src/modals/AssetAndChainSelectorModal/AssetAndChainSelectorModalSearchInput.tsx
```diff
@@ -65,6 +69,7 @@ export const AssetAndChainSelectorModalSearchInput = ({
 
       <StyledSearchInput
         ref={inputRef}
+        autoFocus
```
toddkao, line 72: I really don't know why but adding this attribute seems necessary to not autoFocus on mobile

codingki (approved):
works as expected, lgtm

PR #581, opened 2024-12-17
[FRE-1308] Mobile evm wallet

codingki:


ericHgorski:
Noticed a few UX quirks (weird behavior between FaceID and cross-app prompts) but think its good enough for now. 
another think I noticed: rejecting a tx in mobile doesn't throw error in the app

ericHgorski (approved):
LGTM

PR #589, opened 2024-12-17
Add sentry

toddkao:
Add sentry integration
Add session replay, auto recording when error occurs
Add messages for wallet connected/wallet disconnected


packages/widget/src/state/swapExecutionPage.ts
```diff
@@ -96,6 +97,7 @@ export const setSwapExecutionStateAtom = atom(null, (get, set) => {
       }
     },
     onTransactionBroadcast: async (txInfo) => {
+      setUser({ id: txInfo?.txHash });
```
ericHgorski, line 100: did we want to set this to the user address?

packages/widget/src/pages/SwapPage/SwapPage.tsx
```diff
@@ -304,6 +305,7 @@ export const SwapPage = () => {
           }
           setChainAddresses({});
           setCurrentPage(Routes.SwapExecutionPage);
+          setUser({ username: sourceAccount?.address });
```
toddkao, line 308: @ericHgorski we have two calls to setUser, the one to set address is here

packages/widget/src/state/swapExecutionPage.ts
toddkao: yeah, we should keep both, one is setting id and one is setting username so both will be kept

packages/widget/src/pages/SwapPage/SwapPage.tsx
```diff
@@ -304,6 +305,7 @@ export const SwapPage = () => {
...
           setSwapExecutionState();
```
codingki, line 309: lets also capture the route here

toddkao: how should i capture the route?

codingki: not blocking the merge, we can use this https://docs.sentry.io/platforms/javascript/enriching-events/context/

codingki (approved):

packages/widget/src/pages/SwapPage/SwapPage.tsx
toddkao: hmm, ok I'll test tags and contexts in a separate PR to decide which one to use

https://docs.sentry.io/platforms/javascript/enriching-events/tags/
https://docs.sentry.io/platforms/javascript/enriching-events/context/

PR #592, opened 2024-12-17
improve example

ericHgorski:


codingki (approved):
lgtm, we support other chain other than mainnet now nice!

PR #597, opened 2024-12-18
[FRE-1314] fix: filter available wallets in RenderWalletList component

ericHgorski:


codingki:
also found some unnecessary hook deps

ericHgorski:
> Not working for solana wallets image

This is as intended if only one wallet option is available then we should all available options. I could refine the code so this is only the case with walletconnect and not any length === 1 available wallet list. Its a design question if we're ok with this modal having just a single option.

codingki:
Ok nice, lgtm if the hook deps fixed

packages/widget/src/components/RenderWalletList.tsx
```diff
@@ -106,78 +114,62 @@ export const RenderWalletList = ({
...
         />
       );
     },
     [connectMutation]
   );
 
   const height = useMemo(() => {
-    return Math.min(530, walletList.length * (ITEM_HEIGHT + ITEM_GAP));
+    return Math.min(530, displayWallets.length * (ITEM_HEIGHT + ITEM_GAP));
   }, [walletList]);
```
codingki, line 170:
```suggestion
  }, [displayWallets]);
```

codingki (approved):
merge when eslint deps fixed

PR #602, opened 2024-12-19
Kiki/fre 1321 keplr in app browser issues

codingki:


packages/widget/src/modals/AssetAndChainSelectorModal/AssetAndChainSelectorModalSearchInput.tsx
```diff
@@ -93,6 +93,7 @@ export const AssetAndChainSelectorModalSearchInput = ({
 const StyledLeftArrowIcon = styled(LeftArrowIcon)`
   width: 25px;
   opacity: 0.5;
+  transform: rotate(180deg);
 `;
```
toddkao, line 97: is the `LeftArrowIcon` not a left arrow? I'm a bit confused on this one, we should have a `RightArrowIcon` and `LeftArrowIcon` so this shouldn't be necessary lol, or are we trying to do something weird like rotate it depending on if it's mobile or not?

codingki: yeah it's not rotating on mobile, not sure why but this fixed it

toddkao (approved):
other than the weird arrow rotate styling, looks good

packages/widget/src/modals/AssetAndChainSelectorModal/AssetAndChainSelectorModalSearchInput.tsx
toddkao: can we try moving this rotation inside of ArrowIcon.tsx ?

toddkao: like try updating it so we do rotate(180deg) instead of rotate(180) in there

PR #607, opened 2024-12-20
fix: update SetAddressModal to use theme from useTheme and improve wi…

ericHgorski:
…thdrawal warning display


ericHgorski:
Screenshot 2024-12-20 at 5 21 56 PM

PR #621, opened 2025-01-03
Fix failed to covert value to css stylesheet

toddkao:
Converting `globalStyles` to a `CSSStyleSheet` object before passing it to `react-shadow-scope` seems to fix this issue (instead of passing it directly as a string/template literal)

https://github.com/jonathandewitt-dev/react-shadow-scope?tab=readme-ov-file#constructed-style-sheets


ericHgorski (approved):
LGTM!

PR #627, opened 2025-01-07
[FRE-1370] Fix route preference selector

toddkao:
Fix route preference selector

### DIFF
--- a/packages/widget/src/devMode/loadWidget.tsx
+++ b/packages/widget/src/devMode/loadWidget.tsx
@@ -40,12 +40,6 @@ const DevMode = () => {
         }}>
           <Widget
             theme={theme}
-            settings={{
-              slippage: 5,
-            }}
-            routeConfig={{
-              goFast: true,
-            }}
           />
         </div>
       </Row>
--- a/.changeset/cool-boats-attack.md
+++ b/.changeset/cool-boats-attack.md
@@ -1,0 +1,5 @@
+---
+'@skip-go/widget': patch
+---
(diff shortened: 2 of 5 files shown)


packages/widget/src/devMode/loadWidget.tsx
```diff
@@ -40,12 +40,6 @@ const DevMode = () => {
         }}>
           <Widget
             theme={theme}
-            settings={{
-              slippage: 5,
-            }}
-            routeConfig={{
-              goFast: true,
-            }}
```
toddkao, line 48: these props would cause the slippage/routePreference to always default to 5 and goFast: true in dev mode (more specifically, in http://localhost:5173) but it's unrelated to the bug. I figure we can just remove it so it doesn't confuse us later

toddkao: there's also a question here as to whether this is the intended behavior or if it should only initially default to slippage:5/goFast:true before the user changes the setting at which point itll just save what the user selects going forward?

toddkao: but if we do want to change that we can do it in a seperate PR

ericHgorski: this makes sense!

ericHgorski: I think this behavior is ok for now

ericHgorski (approved):
LGTM!

PR #629, opened 2025-01-08
Update to include prettier-like rules into eslint

toddkao:
should auto apply eslint rules on save


packages/widget/eslint.config.mjs
```diff
@@ -11,24 +12,44 @@ export default tseslint.config(
...
+        {
+          singleQuote: false,
+          trailingComma: "all",
+          tabWidth: 2,
+          semi: true,
+          printWidth: 100,
+          bracketSpacing: true,
+          endOfLine: "auto",
+        },
+      ],
```
toddkao, line 30: these are the "prettier-like" rules added to eslint

ericHgorski (approved):
LGTM

PR #631, opened 2025-01-09
[FRE-1382] Track error in sentry if the transaction fails

toddkao:
Track error in sentry if the transaction fails


ericHgorski (approved):
lgtm

PR #636, opened 2025-01-10
[FRE-1382] Fix skip go is sometimes showing success when status is STATE_COMPLETED_ERROR

toddkao:
image

### DIFF
--- a/packages/widget/src/pages/ErrorPage/ErrorPageTransactionFailed.tsx
+++ b/packages/widget/src/pages/ErrorPage/ErrorPageTransactionFailed.tsx
@@ -14,14 +14,14 @@ import { captureException } from "@sentry/browser";
 import { useEffect } from "react";
 
 export type ErrorPageTransactionFailedProps = {
-  transactionHash: string;
+  txHash: string;
   explorerLink: string;
   onClickContactSupport: () => void;
-  onClickBack: () => void;
+  onClickBack?: () => void;
 };
 
 export const ErrorPageTransactionFailed = ({
-  transactionHash,
+  txHash,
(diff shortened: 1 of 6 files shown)


packages/widget/src/pages/SwapExecutionPage/useBroadcastedTxs.ts
```diff
@@ -51,19 +51,19 @@ export const useBroadcastedTxsStatus = ({
         );
       });
 
-      const _isAllTxSuccess = transferEvents.every((tx) => {
-        return tx.status === "completed";
-      });
-
```
toddkao, line 57: `_isAllTxSuccess` is the root of the issue, sometimes transferEvents all succeed even though the overall swap fails

packages/widget/src/pages/ErrorPage/ErrorPageTransactionFailed.tsx
```diff
@@ -14,14 +14,14 @@ import { captureException } from "@sentry/browser";
 import { useEffect } from "react";
 
 export type ErrorPageTransactionFailedProps = {
-  transactionHash: string;
+  txHash: string;
```
toddkao, line 17: to be consistent

toddkao:
I think I can just look for an error in any TX instead of just the last one

packages/widget/src/pages/ErrorPage/ErrorPageTransactionFailed.tsx
ericHgorski: Love this change

codingki (approved):

PR #650, opened 2025-01-13
Fix showing SetAddressModal if chain id isn't supported by the wallet

toddkao:
Fix showing SetAddressModal if chain id isn't supported by the wallet


codingki (approved):
lgtm

PR #669, opened 2025-01-16
refactor: extract source status logic into SwapExecutionPage

ericHgorski:


packages/widget/src/pages/SwapExecutionPage/SwapExecutionPage.tsx
```diff
@@ -138,6 +158,7 @@ export const SwapExecutionPage = () => {
         operations={clientOperations}
         statusData={statusData}
         swapExecutionState={swapExecutionState}
+        sourceStatus={sourceStatus}
```
toddkao, line 161: I would rename this to be like "firstOperationStatus"

toddkao (commented):
I kinda feel like it's weird to have the firstOperationStatus hoisted to a higher level / not have all of the status logic in the same place

ericHgorski:
> I feel like it's weird to have the firstOperationStatus hoisted to a higher level / not have all of the status logic in the same place but I don't feel that strongly about it

Not sure I follow! Isnt this in one place now instead of 2?

toddkao:
> > I feel like it's weird to have the firstOperationStatus hoisted to a higher level / not have all of the status logic in the same place but I don't feel that strongly about it
> 
> Not sure I follow! Isnt this in one place now instead of 2?

I just mean that the status of the other operations aren't also in SwapExecutionPage

ericHgorski:
> > > I feel like it's weird to have the firstOperationStatus hoisted to a higher level / not have all of the status logic in the same place but I don't feel that strongly about it
> > 
> > 
> > Not sure I follow! Isnt this in one place now instead of 2?
> 
> I just mean that the status of the other operations aren't also in SwapExecutionPage

I do feel like ideally this would be the case. but think this is still step in the right direction

toddkao:
> > > > I feel like it's weird to have the firstOperationStatus hoisted to a higher level / not have all of the status logic in the same place but I don't feel that strongly about it
> > > 
> > > 
> > > Not sure I follow! Isnt this in one place now instead of 2?
> > 
> > 
> > I just mean that the status of the other operations aren't also in SwapExecutionPage
> 
> I do feel like ideally this would be the case. but think this is still step in the right direction

Yeah I agree if we can simply lift most of the logic up to `SwapExecutionPage` it would be ideal

toddkao (approved):

PR #675, opened 2025-01-17
FRE-1393 Sp/settingswheel

entityleak:
https://linear.app/skip/issue/FRE-1393/increase-size-and-contrast-of-settings-wheel-icon

+ settings bottom drawer spacing fixes
+ settings changed indicator not being right colour for theme in light mode
+ increase height of footer button element per design


ericHgorski:
missing changeset!

codingki (approved):

PR #676, opened 2025-01-17
tunnel to /api/sentry

toddkao:


packages/widget/src/index.tsx
```diff
@@ -16,6 +16,7 @@ import {
 
 init({
   dsn: "https://10ce608bdd1c68a13d3849d6b242333c@o4504768725909504.ingest.us.sentry.io/4508485201231872",
+  tunnel: "/api/sentry",
```
codingki, line 19: should it be `https://go.skip.build/api/sentry`

toddkao: both absolute and relative paths should work, in their example they showed a relative url like "/tunnel"

toddkao: i guess using https://go.skip.build/api/sentry would work better if we're testing locally

toddkao: i can update

codingki: yeah better use `https://go.skip.build/api/sentry` because this widget can be used outside go.skip.build

codingki (approved):

PR #686, opened 2025-01-20
Enable gradient colors

toddkao:
feature request from sei to enable gradient colors:

1. Changed background-image to background
2. Added brandTextColor to theme to enable manually passing a brandTextColor
3. Updated icons in MainButton component to have a "cutout" so it works with gradient backgrounds

image


codingki (approved):
lgtm!

PR #689, opened 2025-01-21
Add more data to tx callbacks

toddkao:


codingki (approved):
lgtm, let's add it to the docs as well

PR #693, opened 2025-01-22
fix balance infinite spinner bug

toddkao:
tanstack query v5 docs show [here](https://tanstack.com/query/latest/docs/framework/react/guides/disabling-queries#isloading-previously-isinitialloading) that isLoading (isInitialLoading in v4) is derived from isPending && isFetching, so we can simply use that instead of isLoading and it will work in both v4 and v5

isLoading in v4 is bugged and will be true if the query is disabled but never actually fetched/queried

### DIFF
--- a/packages/widget/src/hooks/useGetSourceBalance.ts
+++ b/packages/widget/src/hooks/useGetSourceBalance.ts
@@ -17,6 +17,8 @@ export const useGetSourceBalance = () => {
     address: sourceAccount?.address,
   });
 
+  console.log(isFetched || isLoading || cw20Balance.isLoading);
+
   const data = useMemo(() => {
     if (!sourceAsset || !sourceAccount || !skipBalances) return;
     const { chainID, denom } = sourceAsset;
--- a/packages/widget/src/hooks/useGetSourceBalance.ts
+++ b/packages/widget/src/hooks/useGetSourceBalance.ts
@@ -17,7 +17,8 @@ export const useGetSourceBalance = () => {
     address: sourceAccount?.address,
   });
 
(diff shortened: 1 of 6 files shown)


packages/widget/src/hooks/useGetSourceBalance.ts
```diff
@@ -17,6 +17,9 @@ export const useGetSourceBalance = () => {
     address: sourceAccount?.address,
   });
 
+  // this is to support both tanstack query v4/v5
+  const cw20BalanceIsLoading = cw20Balance.isPending && cw20Balance.isFetching;
```
codingki, line 21: I think the correct logic is `cw20Balance.isPending || cw20Balance.isFetching;`

codingki:
I think we should change all `isLoading` to this

packages/widget/src/hooks/useGetSourceBalance.ts
toddkao: I took the logic from https://tanstack.com/query/latest/docs/framework/react/guides/disabling-queries#isloading-previously-isinitialloading

toddkao: they said it's cw20Balance.isPending && cw20Balance.isFetching;

toddkao:
> I think we should change all react query's `isLoading` to this

i wish there was an easy way to do this once and not have to duplicate it 🤔

packages/widget/src/hooks/useGetSourceBalance.ts
codingki: I see, approved

codingki (approved):

PR #710, opened 2025-01-23
Simplify search to use substring search

toddkao:


ericHgorski (approved):
needs changeset but lgtm

PR #717, opened 2025-01-25
Fix: dont override routeconfig in mobile hook

ericHgorski:


codingki:
because this is a hotfix, lets target to main

toddkao (approved):
changeset, and also there's an eslint warning but it's a nit

ericHgorski:
> because this is a hotfix, lets target to main

Going to get this out as part of regular release instead.

PR #722, opened 2025-01-27
[FRE-1446] elys cosmjs v0320 support in the widget

codingki:


docs/widget/getting-started.mdx
```diff
@@ -35,7 +35,7 @@ If you're using `yarn` (or another package manager that doesn't install peer dep
 you may need to install these peer dependencies as well:
 
 ```bash
-  yarn add @tanstack/react-query graz react react-dom viem wagmi
+  yarn add @tanstack/react-query viem wagmi
```
toddkao, line 38: let's update the ```shell Yarn example above to also be `yarn add @tanstack/react-query viem wagmi`

toddkao (approved):
the yarn docs stuff is nit, but i do think we should revert the client lib workspace dep though

PR #733, opened 2025-01-29
[FRE-1448] refactor: update imports to use ESM module paths and standardize quotes

ericHgorski:
reduced bundle size by 4mb by removing cjs imports from @injectivelabs dependency


ericHgorski:
duplicate #734

PR #736, opened 2025-01-29
fix bug with evm recovery address not working on mobile

toddkao:
image

### DIFF
--- a/packages/widget/src/hooks/useAutoSetAddress.ts
+++ b/packages/widget/src/hooks/useAutoSetAddress.ts
@@ -80,6 +80,13 @@ export const useAutoSetAddress = () => {
           }));
           return;
         }
+        const showSetAddressModal = () => {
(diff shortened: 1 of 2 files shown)


packages/widget/src/hooks/useAutoSetAddress.ts
```diff
@@ -181,10 +173,7 @@ export const useAutoSetAddress = () => {
               }
               if (!wallet) {
                 if (!openModal) return;
-                NiceModal.show(Modals.SetAddressModal, {
-                  signRequired: isSignRequired,
-                  chainId: chainID,
-                });
```
toddkao, line 187: forgot to pass index here :( 
I made a function to handle this so we dont have to worry about remembering to pass index through

ericHgorski (approved):
LGTM

PR #739, opened 2025-01-30
[FRE-1459, FRE-1460, FRE-1459]

ericHgorski:


packages/widget/src/components/Modal.tsx
```diff
@@ -172,7 +172,6 @@ const StyledOverlay = styled(Dialog.Overlay)<{
   bottom: 0;
   display: grid;
   place-items: center;
-  overflow-y: auto;
```
ericHgorski, line 175: This solved the rounded corners issue and didnt find it broke anythign else, but lmk if this feels important to maintain. @toddkao

toddkao: looks good

toddkao (approved):

PR #753, opened 2025-02-04
Disable replay compression as workaround fix for sentry requests 400'ing

toddkao:
thread about issue here -> https://github.com/getsentry/sentry-javascript/issues/7302

I found the workaround in here. The suggested fix using the tunnelRoute option didn't work for me, I suspect because we aren't actually initializing sentry within nextjs but within the widget.

I also checked the difference between compressed and uncompressed data sent from the client and it's around like 30 bytes more per network request, 70bytes vs 100bytes or so

### DIFF
--- a/.changeset/violet-dingos-do.md
+++ b/.changeset/violet-dingos-do.md
@@ -1,0 +1,5 @@
+---
+"@skip-go/widget": patch
+---
+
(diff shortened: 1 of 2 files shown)


ericHgorski (approved):
LGTM

PR #756, opened 2025-02-05
Transition utils and hover states

entityleak:


ericHgorski:
Hover state on `select asset` is different than on `MainButton`

entityleak:
Should be resolved now

ericHgorski (approved):
LGTM

PR #759, opened 2025-02-06
widget: add support for taddrs on penumbra

erwanor:
This PR:
- restores USDC flows from noble using transparent addresses
- otherwise defaults to ephemeral addresses

I wasn't able to test the PR end-to-end, because trying to generate a route leads me to this error:

Screenshot 2025-02-05 at 11 42 13 PM

Which I couldn't trace the source of. I think this might coming from the solver backend directly? unsure. If you tell me where to look I should be able to test this.


ericHgorski:
@NotJeremyLiu to test end-to-end with local API

erwanor:
@NotJeremyLiu opened #770, if you test this PR you will learn that the current address validation in the client packages does not handle Penumbra. Hopefully this saves some time, happy to test things here with you if that's helpful.

codingki (approved):
I'll pass `sourceChainID` after merge, I test this flow works fine. Thanks

PR #764, opened 2025-02-07
Add mobile devtool and remote debugger to nextjs example

toddkao:
Adds mobile devtool and remote debugger to nextjs example for mobile testing/debugging

image

image


ericHgorski (approved):
LGTM

PR #770, opened 2025-02-08
Fix address validation to work with Penumbra chains

erwanor:
Trying to make transfers to Penumbra from anyhere in the interchain fails with this error on the main Skip app (not in the widget, curiously):
Screenshot 2025-02-07 at 7 24 44 PM

After digging, I learned that this started after this commit https://github.com/skip-mev/skip-go/commit/8be4ad9b66511041d52d6c27e526934d1f8a55fc added address validation.

The source of the error is that Penumbra uses `bech32m` addresses.

I wish we could stop there, I would have pushed this simple change: https://github.com/skip-mev/skip-go/commit/d512a6e8ea0634c35aa4a7b1ee4aaac99ba034c3 and carried on. Unfortunately, the diff has to be bigger because of interop issues with Noble mainnet.

Some context:

Penumbra addresses contain encrypted metadata. For a variety of reasons, this means that addresses must be encoded using `bech32m`.

1. Noble middleware first broke transfer with Penumbra chains, so we created a compat `bech32` compatible address to make it work, with prefix `penumbracompat1`
2. Unfortunately, a second breakage happened and transfers with both the compat and ordinary privacy addresses broke, so we invented a transparent address format which is a `bech32` address with a `tpenumbra` prefix (#759)
3. Noble has deployed a good fix on their testnet, it should hit mainnet at the end of the month

In the meantime, this PR (along with #759) restores Skip functionality for Penumbra users. Once the Noble fix lands, I would be happy to straighten up the code here and in the fetch address logic.

### DIFF
--- a/package.json
+++ b/package.json
@@ -92,6 +92,7 @@
   },
   "packageManager": "yarn@3.2.0",
   "dependencies": {
-    "@changesets/cli": "^2.24.4"
+    "@changesets/cli": "^2.24.4",
+    "bech32": "^2.0.0"
   }
 }
--- a/packages/client/src/client.ts
+++ b/packages/client/src/client.ts
@@ -1867,13 +1867,20 @@ export class SkipClient {
       switch (chain?.chainType) {
         case types.ChainType.Cosmos:
           try {
-            const prefix = chain.chainID.includes("penumbra")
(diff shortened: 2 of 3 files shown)


codingki (approved):
the dependencies should live in client's package.json. I'll fix this after merge. thanks!

PR #771, opened 2025-02-08
Enhance getMainnetAndTestnetChains to include EVM and SVM options

ericHgorski:


NotJeremyLiu (approved):

codingki:
currently we don't need this, afaik this only for cosmos chains

ericHgorski:
> currently we don't need this, afaik this only for cosmos chains

We have validation logic at top of `executeRoute`. It is strange however because what was previously used here did not include EVM/SVM chains. 

``` async executeRoute(options: clientTypes.ExecuteRouteOptions) {
    const { route, userAddresses, beforeMsg, afterMsg } = options;

    let addressList: string[] = [];
    let i = 0;
    for (let j = 0; j < userAddresses.length; j++) {
      if (route.requiredChainAddresses[i] !== userAddresses[j]?.chainID) {
        i = j;
        continue;
      }
      addressList.push(userAddresses[j]!.address!);
      i++;
    }

    if (addressList.length !== route.requiredChainAddresses.length) {
      addressList = userAddresses.map((x) => x.address);
    }

    const validLength =
      addressList.length === route.requiredChainAddresses.length ||
      addressList.length === route.chainIDs.length;

    if (!validLength) {
      raise("executeRoute error: invalid address list");
    }

    const isUserAddressesValid =
      await this.validateUserAddresses(userAddresses);

    if (!isUserAddressesValid) {
      raise("executeRoute error: invalid user addresses");
    }
    ```

codingki (approved):

PR #789, opened 2025-02-12
[FRE-1479] Add Go Fast warning handling and related components

ericHgorski:
Add Go Fast Warning
image

Updated Footer
image

Updated settings drawer to look the same on desktop as it did on mobile:
image

Rejected Auth colors updated:

image

Example with multi sig
image


ericHgorski:
cc @toddkao for PR handoff

PR #797, opened 2025-02-19
Fix premature EOF based on the code in stridejs's repo

toddkao:
Fix premature EOF based on the code in stridejs's repo, primarily by replacing `_m0.Reader` by `BinaryWriter` and replacing Long type with bigint

https://github.com/Stride-Labs/stridejs/blob/main/src/parser.ts#L22-L38

https://github.com/Stride-Labs/stridejs/blob/48ba444d2bd1441cb0d47de4b96b3c58ed568ca3/src/codegen/stride/vesting/vesting.ts#L341-L360

### DIFF
--- a/.changeset/cold-balloons-buy.md
+++ b/.changeset/cold-balloons-buy.md
@@ -1,0 +1,5 @@
+---
+"@skip-go/client": patch
+---
+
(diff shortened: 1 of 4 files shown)


codingki (approved):
lgtm

PR #798, opened 2025-02-19
Add countdownTimer

toddkao:
https://github.com/user-attachments/assets/8b0f4a64-82ed-4b41-86eb-4ca898e1e634

### DIFF
--- a/packages/widget/src/pages/SwapExecutionPage/SwapExecutionButton.tsx
+++ b/packages/widget/src/pages/SwapExecutionPage/SwapExecutionButton.tsx
@@ -50,20 +50,20 @@ export const SwapExecutionButton: React.FC<SwapExecutionButtonProps> = ({
         swapExecutionState,
       )
     ) {
-      const timer = createCountdownTimer({
+      const countdownTimer = createCountdownTimer({
         duration: estimatedDurationSeconds,
         onUpdate: (remainingTime) => {
           setCountdown(parseInt(remainingTime.toString()));
         },
       });
-      setTimer(timer);
-      timer.startCountdown();
+      setTimer(countdownTimer);
+      countdownTimer.startCountdown();
     }
 
     return () => {
       timer?.stopCountdown();
     };
-  }, [countdown, route?.estimatedRouteDurationSeconds, swapExecutionState, timer]);
+  }, [route?.estimatedRouteDurationSeconds, swapExecutionState, timer]);
 
   const theme = useTheme();
   const setError = useSetAtom(errorAtom);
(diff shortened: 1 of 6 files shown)


packages/widget/src/utils/countdownTimer.ts
```diff
@@ -0,0 +1,44 @@
...
+    start = null;
+    animationFrameId = requestAnimationFrame(step);
+  }
+
+  function stopCountdown() {
+    if (animationFrameId !== null) {
+      cancelAnimationFrame(animationFrameId);
+      animationFrameId = null;
+    }
+  }
```
toddkao, line 41: in my experience using rAf for timers is better for several reasons

toddkao:
hmm i just realized if it's showing in minutes we don't really see the countdown timer until a full minute passes

ericHgorski:
> hmm i just realized if it's showing in minutes we don't really see the countdown timer until a full minute passes, perhaps we should instead show in "mm:ss" format

Agree with this

entityleak:
Is this timer accurate? In the video is completes at 20 second mark.

For times longer than 1 minute, we should not show seconds, just show the minutes tick down. (15m, 14m, 13m) . (It's not really important to show to users that there's 14:12 remaning)

 Once it goes sub-one-minute we should show the seconds counting down.  (not MM:SS, which takes up too much space)

ericHgorski:
> }

Ok @toddkao to revert mm:ss change.

toddkao:
> Is this timer accurate? In the video is completes at 20 second mark.
> 
> For times longer than 1 minute, we should not show seconds, just show the minutes tick down. (15m, 14m, 13m) . (It's not really important to show to users that there's 14:12 remaning)
> 
> Once it goes sub-one-minute we should show the seconds counting down. (59s, 58s, 57s) (not MM:SS, which takes up too much space)

the timer is accurate, the estimate of 30 seconds however isn't it's a rough estimate provided by the API which I understand is a list of estimate times based on the operations in a route added together

packages/widget/src/pages/SwapExecutionPage/SwapExecutionButton.tsx
```diff
@@ -33,6 +35,36 @@ export const SwapExecutionButton: React.FC<SwapExecutionButtonProps> = ({
...
+        },
+      });
+      setTimer(countdownTimer);
+      countdownTimer.startCountdown();
+    }
+
+    return () => {
+      timer?.stopCountdown();
+    };
+  }, [route?.estimatedRouteDurationSeconds, swapExecutionState, timer]);
```
codingki, line 66: can we extract this into a custom hook `useCountdown`
```
const {countdown} = useCountdown({
estimatedRouteDurationSeconds, 
swapExecutionState
})
```

toddkao: done

codingki (approved):
lgtm

PR #800, opened 2025-02-20
Always update sourceWallet if it's undefined when connecting wallet

toddkao:


packages/widget/src/hooks/useCreateCosmosWallets.tsx
```diff
@@ -164,16 +156,17 @@ export const useCreateCosmosWallets = () => {
...
-                throw new Error("always prompt wallet connection");
-              }
-              if (!chainId) {
-                throw new Error("no chain id");
+              const getAddressWithoutConnectingWallet =
+                cosmosWallet && currentWallet !== wallet && !signRequired && chainId;
+
+              if (getAddressWithoutConnectingWallet) {
+                const address = (await getWallet(wallet).getKey(chainId)).bech32Address;
+                return { address };
```
toddkao, line 164: @codingki is there a way to achieve this behavior for EVM/SVM ?

codingki: for svm I believe it works, but we can't test this because there are no routes svm -> svm

toddkao: ah ok, i guess we don't need to worry about those for now 😆

codingki (approved):

PR #802, opened 2025-02-20
fix ability for recovery addresses to be auto updated when wallet is changed

toddkao:
- Split `useKeepWalletStateSynced` part of `useGetAccount` logic and moved it to a separate hook
- Removed other places we update source wallet, and just rely on `useKeepWalletStateSynced`
- Stop using `useQuery` for `useAutoSetAddresses`, instead only call `connectRequiredChains()` when connected wallets have changed

https://github.com/user-attachments/assets/f599219f-83d3-4a65-99b0-d5ee222ef740

### DIFF
--- a/packages/widget/src/state/wallets.ts
+++ b/packages/widget/src/state/wallets.ts
@@ -38,6 +38,7 @@ export type MinimalWallet = {
 };
 
 type WalletState = {
+  id?: string;
   walletName: string;
   chainType: string;
   logo?: string;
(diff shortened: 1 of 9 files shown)


packages/widget/src/state/wallets.ts
```diff
@@ -38,6 +38,7 @@ export type MinimalWallet = {
 };
 
 type WalletState = {
+  id?: string;
```
toddkao, line 41: we just need some kind of identifier, I think using one of the addresses associated to the wallet should be fine?

PR #807, opened 2025-02-24
Fix eof error

toddkao:


.changeset/cold-balloons-buy.md
```diff
@@ -0,0 +1,5 @@
+---
+"@skip-go/client": patch
```
codingki, line 2: I think you should patch widget too

codingki (approved):
update patch changeset for widget

.changeset/cold-balloons-buy.md
toddkao: good catch

PR #808, opened 2025-02-24
Fix solana wallet detection in useKeepWalletStateSynced

toddkao:


packages/widget/src/hooks/useKeepWalletStateSynced.ts
```diff
@@ -16,7 +16,9 @@ export const useKeepWalletStateSynced = () => {
     multiChain: true,
   });
 
-  const solanaWallet = solanaWallets.find((w) => w.name === wallet.svm?.walletName);
+  const solanaWallet = solanaWallets.find(
+    (wallet) => wallet.readyState === WalletReadyState.Installed,
+  );
```
codingki, line 21: I don't think this is correct, why we store a wallet that being installed?

toddkao: the idea is that we are updating the app state based on the wallet that is already connected, ("installed" i believe is the state for connected?)

ericHgorski: I'd imagine that just means its installed as an extension?

toddkao: ah yeah you're right, I was looking at the wrong state

codingki (approved):
lgtm

codingki (approved):

PR #812, opened 2025-02-25
docs: sync to cli README, add transfer back commands, cleaning

olliegilbey:
DOCS:
This PR updates the docs to align with the changes to the README of the CLI tool, ahead of stripping out the CLI README so that these docs become the source of truth.

### DIFF
--- a/docs/eureka/devnet-testing-guide.mdx
+++ b/docs/eureka/devnet-testing-guide.mdx
@@ -1,11 +1,13 @@
 ---
-title: "Devnet Testing Guide"
+title: "Eureka CLI Devnet Testing Guide"
+id: eureka-cli
 description: "This document describes how to test IBC Eureka on Devnet"
 ---
 
 # Background
 
-IBC Eureka is the official native implementation of IBC v2 that leverages the Cosmos Hub and Interchain Labs contracts, relayers, and infrastructure to enable seamless cross-chain user and liquidity mobility.
+IBC Eureka is the official native implementation of IBC v2
+It leverages the Cosmos Hub and Interchain Labs contracts, relayers, and infrastructure to enable seamless cross-chain user and liquidity mobility.
 
 The following guide explains how to test IBC Eureka's Devnet deployment via a CLI tool developed to interact with IBC Eureka, enabling trust-minimized token transfers between Ethereum and Cosmos chains.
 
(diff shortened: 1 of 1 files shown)


docs/eureka/devnet-testing-guide.mdx
```diff
@@ -16,7 +18,7 @@ The following guide explains how to test IBC Eureka's Devnet deployment via a CL
 - Access to Ethereum and Cosmos endpoints
 - Private keys for both chains
 - Sepolia ETH for gas fees on the Ethereum Sepolia Testnet
-- An ERC20 token on Ethereum Sepolia Testnet
+- An ERC20 token on Ethereum Sepolia Testnet such as our [$TNERC](https://sepolia.etherscan.io/token/0xa4ff49eb6e2ea77d7d8091f1501385078642603f)
```
NotJeremyLiu, line 21: What's the value of adding this ERC20 token if the tester doesn't have access to it? Or can they mint it themselves which is why we're linking it?

docs/eureka/devnet-testing-guide.mdx
```diff
@@ -128,6 +129,76 @@ Example Output:
...
+transfer/08-wasm-0/0xa4ff49eb6e2ea77d7d8091f1501385078642603f: 1
+```
+
+You can make an equivalent query using `gaiad` with:
+
+```bash
+gaiad query bank balances cosmos1u5d4hk8294fs9pq556jxmlju2ceh4jmurcpfv7 --chain-id=highway-dev-1 --node=https://eureka-devnet-node-01-rpc.dev.skip.build:443 --output json
+```
+
+### Transfer Tokens from Ethereum to Cosmos (Back Again!)
```
NotJeremyLiu, line 160: I think this should exist above the balance queries, and balance queries last. To keep the one direction and the other commands together for better organization

docs/eureka/devnet-testing-guide.mdx
```diff
@@ -1,11 +1,13 @@
 ---
-title: "Devnet Testing Guide"
+title: "Eureka CLI Devnet Testing Guide"
```
NotJeremyLiu, line 2: I think this is too long on the side menu so just keeping it "Devnet Testing Guide" prob better
CleanShot 2025-02-25 at 07 56 27

docs/eureka/devnet-testing-guide.mdx
olliegilbey: This was thinking that they could ask for some/we send some/we give wallets with some.
Since we haven't - can scrap and have them create themselves.

docs/eureka/devnet-testing-guide.mdx
olliegilbey: agre :)

docs/eureka/devnet-testing-guide.mdx
NotJeremyLiu: Yeah I'm p indifferent on keeping the balance command instructions for cosmos or not, only main thing for me here is that the transfer commands should all be groped together

NotJeremyLiu (approved):

PR #814, opened 2025-02-25
[FRE-1498, FRE-1500, FRE-1501] manually set walletAtoms to undefined when disconnecting, fix getAddressWithoutConnectingWallet logic

toddkao:


codingki (approved):
lgtm

PR #832, opened 2025-02-27
update currentCosmosId to be based on the first address in the account instead of hard-coded to cosmoshub-4

toddkao:


codingki (approved):

codingki:
changeset

PR #837, opened 2025-02-28
Fix solana signers

toddkao:


packages/widget/src/hooks/useCreateEvmWallets.tsx
```diff
@@ -20,7 +20,7 @@ export const useCreateEvmWallets = () => {
   const { data: chains } = useAtomValue(skipChainsAtom);
   const { data: assets } = useAtomValue(skipAssetsAtom);
   const [sourceAsset, setSourceAsset] = useAtom(sourceAssetAtom);
-  const [evmWallet, setEvmWallet] = useSetAtom(evmWalletAtom);
+  const [evmWallet, setEvmWallet] = useAtom(evmWalletAtom);
```
toddkao, line 23: i forgot to update useSetAtom to useAtom in my previous PR https://github.com/skip-mev/skip-go/pull/836

packages/widget/src/state/skipClient.ts
```diff
@@ -25,11 +25,15 @@ export const skipClientConfigAtom = atom<SkipClientOptions>({
 
 export const themeAtom = atom<Theme>(defaultTheme);
 
+export const solanaWalletsAtom = atom<Wallet[]>();
```
codingki, line 28: I don't think this no need to be an atom, can you elaborate why we need this to be an atom?

toddkao: main issue is we can't get the wallets from the context of `SolanaProvider` in jotai besides doing something like this

toddkao: the alternative obviously would be to just stop using the SolanaProvider

codingki: let's stop using SolanaProvider and revert to what we were doing before. too op if we using only `useWallet` hooks from the solana react package. because SolanaProvider or context is useless here

codingki (changes requested):
let's stop using SolanaProvider and revert to what we were doing before. too op if we using only useWallet hooks from the solana react package. because SolanaProvider or context is useless here

codingki (approved):
lgtm

PR #846, opened 2025-03-03
feat: history page add transfer asset release info

codingki:


packages/widget/src/pages/TransactionHistoryPage/TransactionHistoryPage.tsx
```diff
@@ -37,7 +35,7 @@ export const TransactionHistoryPage = () => {
         <VirtualList
           key={txHistory.length}
           listItems={historyList}
-          height={isMobileScreenSize ? 0 : 262}
+          height={262}
```
toddkao, line 38: I remember we specifically added this so that mobile would not have a limited height, not sure why we changed our minds on this, but I'm ok with it either way

toddkao (approved):

PR #864, opened 2025-03-05
[FRE-1543] read and add a low info warning to the widget

codingki:
renamed "trade warning" to "bad price warning"
image


toddkao (approved):

toddkao:
i wonder if we should show the input/output values (in crypto denomination) on this page for them to review

PR #868, opened 2025-03-06
eureka transfer support

dhfang:


toddkao (commented):
should we update `SwapSettingsDrawer.tsx` if there's fees associated with EurekeaTransfer? 
I'm also curious if our getOperationDetailsAndType function will work properly with eurekaTransfers or not.. 
Besides that everything looks good

codingki:
I think the fees here is smart relay fee, which already covered

> should we update `SwapSettingsDrawer.tsx` if there's fees associated with EurekaTransfer? I'm also curious if our `getOperationDetailsAndType` function will work properly with eurekaTransfers or not.. Besides that everything looks good

codingki:
> I'm also curious if our getOperationDetailsAndType function will work properly with eurekaTransfers or not..
Besides that everything looks good

I think these values are we care about
```ts
    fromChainID: string;
    toChainID: string;
    denomIn: string;
    denomOut: string;
    bridgeID: BridgeType;
```

PR #875, opened 2025-03-10
[FRE-1550] Improve execute route

toddkao:
- Update `SkipClient` fields to be static so that multiple instances refer to the same values
- Remove caching middleware
- Preload/cache `assets`, `chains`, `balances`, `stargateClient`
- parallelize `validateGasBalances`


ericHgorski:
Do the cached values take into account query params like chainID? 

e.g. If I queries skipChains for only SVM, and then queried again for all chains, would the cache return just SVM chains?
Same goes for balances and assets.

toddkao:
> Do the cached values take into account query params like chainID?
> 
> e.g. If I queries skipChains for only SVM, and then queried again for all chains, would the cache return just SVM chains? Same goes for balances and assets.

yeah that's a good call out

packages/client/src/client.ts
```diff
@@ -1765,18 +1776,21 @@ export class SkipClient {
       }
       return calculateFee(Math.ceil(parseFloat(estimatedGasAmount)), gasPrice);
     });
-    const feeBalance = await this.balances({
-      chains: {
-        [chainID]: {
-          address: signerAddress,
-          denoms: feeAssets.map((asset) => asset.denom),
+    const feeBalance =
+      SkipClient.skipBalances ??
```
codingki, line 1780: we should check the `feeAssets` balances in `.skipBalances` instead of check if the `.skipBalances` data is available

packages/client/src/client.ts
```diff
@@ -145,36 +183,41 @@ export class SkipClient {
...
       },
-      [options],
+      {} as Record<string, types.Asset[]>,
     );
+
+    if (
+      options.includeEvmAssets &&
+      options.includeSvmAssets &&
+      options.includeCW20Assets
+    ) {
```
codingki, line 203: we only store if these params is true?

codingki: oh I see, it's for avoiding diff params

toddkao: yeah, the thinking is that if you call it without one of them being true it might mean our cache is missing some data.

toddkao: there's probably a better / more sophisticated way to handle this but this is simple and should work for our use case

packages/client/src/client.ts
```diff
@@ -230,6 +273,16 @@ export class SkipClient {
       "/v2/info/balances",
       types.balanceRequestToJSON(request),
     );
+    const responseCamelCase = types.balanceResponseFromJSON(response);
+    const mergedBalances = {
+      ...(this.skipBalances?.chains ?? {}),
+      ...responseCamelCase.chains,
+    };
+    if (mergedBalances) {
```
codingki, line 281: unnecessary if, it will always true

packages/client/src/client.ts
```diff
@@ -87,18 +83,21 @@ export class SkipClient {
...
+  getCosmosSigner?: clientTypes.SignerGetters["getCosmosSigner"];
+  getEVMSigner?: clientTypes.SignerGetters["getEVMSigner"];
+  getSVMSigner?: clientTypes.SignerGetters["getSVMSigner"];
+  chainIDsToAffiliates?: clientTypes.SkipClientOptions["chainIDsToAffiliates"];
+  cumulativeAffiliateFeeBPS?: string = "0";
+
+  private clientOptions: clientTypes.SkipClientOptions;
+  private skipChains?: types.Chain[];
+  private skipAssets?: Record<string, types.Asset[]>;
+  private skipBalances?: types.BalanceResponse;
```
codingki, line 95: this var will have an edge case if the user change the address. we will need an address validation in executeRoute to check the address in skipBalances.chains response is same, if not refetch

codingki: for example: 
wallet a: have osmo on osmosis
wallet b: don't have osmo on osmosis

executeRoute using wallet b fee asset is osmo on osmosis:
it will assume the asset is available

toddkao: I think once you change your address and you fetch the new balance it will update the cache again, so unless they manage to call executeRoute before fetching the balance of wallet b I think it won't cause issues?

codingki: because balances is merged like this https://github.com/skip-mev/skip-go/pull/875/files#diff-1d25041624e23bb1be334f0c7033cf4fadf2801dae21687f77f5511cd5935b92R277-R280

codingki: it will only check the key which is the chain id

toddkao: i think the new one will overwrite the old one

toddkao: its easier to not merge them then, and always just replace the old skipBalances

codingki: image

codingki: yes

toddkao: updated!

packages/client/src/client.ts
```diff
@@ -87,18 +83,21 @@ export class SkipClient {
...
+  private stargateClientByChainId: Record<string, SigningStargateClient> = {};
```
codingki, line 96: let's rename this to `signingStargateClientByChainId`

packages/client/src/cache.ts
```diff
@@ -1,57 +0,0 @@
-import createHash from 'create-hash';
```
codingki, line 1: if we are not using this package, let's remove it from our deps in package.json

packages/client/src/client.ts
```diff
@@ -683,6 +726,52 @@ export class SkipClient {
     }
   }
 
+  async getStargateClient({
```
codingki, line 729:
```suggestion
  async getSigningStargateClient({
```

codingki (approved):
lgtm

packages/client/src/client.ts
```diff
@@ -1765,18 +1833,28 @@ export class SkipClient {
...
     });
-    const feeBalance = await this.balances({
-      chains: {
-        [chainID]: {
-          address: signerAddress,
-          denoms: feeAssets.map((asset) => asset.denom),
-        },
-      },
+
+    const feeAssetFoundInSkipBalances = feeAssets.some((asset) => {
```
codingki, line 1837: should we use `.every` here

codingki: because if it's undefined we should fetch the balance

toddkao: i was originally thinking they dont need all of the feeAssets, if they have one i believe they can pay with whatever asset they have but it complicates things if they have it but it's not enough... and I think in practice we most likely will have all so I think we can update to using every

PR #884, opened 2025-03-11
add transaction reverted error

codingki:


toddkao (approved):
nit: there's an extra empty file useHandleTransactionReverted that seems like should be removed?

PR #886, opened 2025-03-12
[FRE-1553] Add prop to disable shadowDOM

toddkao:
- Add prop to disable shadowDOM
- Update/Fix settings and modals to work properly when shadowDOM is disabled
- Update styling so it is the same regardless of if shadowDOM is enabled/disabled
- Fix flickering animation when modals are opened when shadowDOM is disabled


codingki (approved):
lgtm, please update docs and changeset

PR #891, opened 2025-03-14
Set bech32 decode limits correctly

zmanian:
Updates the bech32 decode limits in client and widget components


codingki (approved):

codingki:
it's should be added here too https://github.com/skip-mev/skip-go/pull/891/files#diff-e52da860d5947d6827611ff24f3a62342201237cc1397c4c7d1c2f828ce06452R33

codingki:
nvm, i read it wrong. lgtm

PR #893, opened 2025-03-14
feat: upgrade mintlify to new docs.json format

olliegilbey:
Migrates mint.json to docs.json as per mintlify suggestion:
https://mintlify.com/docs/settings/global

Migrates Google Tag Manager ID.
Deletes old mint.json

Check this builds correctly on deploy. (Works locally)


NotJeremyLiu (approved):
noiceeee

PR #894, opened 2025-03-14
fix incorrect wallet address display while executing route

codingki:


packages/widget/src/hooks/useAutoSetAddress.ts
```diff
@@ -46,6 +48,7 @@ export const useAutoSetAddress = () => {
 
   const connectRequiredChains = useCallback(
     async (openModal?: boolean) => {
+      if (isPending) return;
```
toddkao, line 51: @codingki i think it might make more sense to check the swapExecutionStateAtom's overallStatus is unconfirmed and returning early if it isnt in the useEffect to avoid calling connectRequiredChains instead of returning early in this function?

codingki: I see, make sense

toddkao (approved):

PR #903, opened 2025-03-17
Update tests to connect to keplr

toddkao:


.github/workflows/tests.yml
```diff
@@ -1,5 +1,10 @@
 name: Tests
-on: [pull_request, push]
+on:
+  pull_request:
+  push:
+    branches:
+      - main
+      - staging
```
toddkao, line 7: this just avoids running tests twice when pushing changes to a PR

codingki (approved):

PR #909, opened 2025-03-18
[FRE-1519] Simplify loading states, fix loading spinners sometimes being out of sync

toddkao:
https://github.com/user-attachments/assets/125b1663-c651-4dfb-b95c-327e6a2cdbbb

### DIFF
--- a/packages/widget/src/pages/SwapExecutionPage/SwapExecutionPageRouteDetailed.tsx
+++ b/packages/widget/src/pages/SwapExecutionPage/SwapExecutionPageRouteDetailed.tsx
@@ -152,9 +152,13 @@ export const SwapExecutionPageRouteDetailed = ({
         : status?.[operation.transferIndex]?.toExplorerLink;
 
       const operationStatus = getOperationStatus(operation);
+      const numberOfTransferEvents = statusData?.transferEvents.length;
+      const latestStatus = status?.[status.length - 1]?.status;
 
       return (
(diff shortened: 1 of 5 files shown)


packages/widget/src/pages/SwapExecutionPage/SwapExecutionPageRouteDetailed.tsx
```diff
@@ -147,9 +152,13 @@ export const SwapExecutionPageRouteDetailed = ({
         : status?.[operation.transferIndex]?.toExplorerLink;
 
       const operationStatus = getOperationStatus(operation);
+      const numberOfTransferEvents = statusData?.transferEvents.length;
+      const latestStatus = status?.[status.length - 1]?.status;
```
codingki, line 156: is this ok to be a key?, because this variable is changing

toddkao: I added these to force it to remount every time these change, that way it will force the loading circles to restart and keep in sync

toddkao: but now that you got me thinking more about it, i can try moving that key to just the loading circle maybe i wonder if that would work too

codingki: I see

codingki (approved):

PR #950, opened 2025-03-26
[FRE-1506] Dynamic Font Size Reduction for Large Numbers

daps94:
Overview
This PR adds automatic text resizing for large numeric values in the widget's input fields. When a number exceeds 14 characters, the font size is automatically reduced to ensure it fits properly in the container without being cut off or overflowing.

Implementation Details
Added a new utility function shouldReduceFontSize() that determines when a number has too many characters
Implemented responsive font sizing in StyledInput component that applies smaller text only when needed
Used styled-component props for type-safe styling logic following codebase conventions
Applied proportional size reduction (38px → 30px on desktop, 30px → 24px on mobile)

### DIFF
--- a/packages/widget/src/pages/SwapPage/SwapPageAssetChainInput.tsx
+++ b/packages/widget/src/pages/SwapPage/SwapPageAssetChainInput.tsx
@@ -172,7 +172,7 @@ export const SwapPageAssetChainInput = ({
           onKeyDown={handleKeyDown}
           disabled={disabled}
           isWaitingToUpdateInputValue={isWaitingToUpdateInputValue}
-          data-large-number={isLargeNumber.toString()}
+          isLargeNumber={isLargeNumber}
         />
         <StyledAssetButton onClick={handleChangeAsset} disabled={disabled} gap={5}>
           {assetDetails?.assetImage && assetDetails.symbol ? (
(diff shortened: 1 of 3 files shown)


packages/widget/src/pages/SwapPage/SwapPageAssetChainInput.tsx
```diff
@@ -170,6 +172,7 @@ export const SwapPageAssetChainInput = ({
           onKeyDown={handleKeyDown}
           disabled={disabled}
           isWaitingToUpdateInputValue={isWaitingToUpdateInputValue}
+          data-large-number={isLargeNumber.toString()}
```
toddkao, line 175: @daps94 

i have nothing against passing data values this way but the typical pattern we use to pass any prop data to `StyledComponents` is the same way we're passing isWaitingToUpdateInputValue, by specifying the prop like this:

```ts
const StyledInput = styled.input<{
  isWaitingToUpdateInputValue?: boolean;
  isLargeNumber?: boolean;
}>`
```

and then using it like this:

```ts
${(props) => props.isLargeNumber};
```

unless there's a reason we need it in the dom/on the html element specifically I'd prefer if we just stay consistent 

this allows us to have type information as well as pass more complex objects if necessary

toddkao: for adding changeset, run `npx changeset` in the root of the app and follow prompts (spacebar to select enter to confirm) and then type a changelog for your changes

toddkao (approved):

packages/widget/src/pages/SwapPage/SwapPageAssetChainInput.tsx
daps94: Thanks @toddkao i've refactored the implementation and added a change log. This should be good to go.

PR #969, opened 2025-03-28
check chainId getAssets

codingki:


packages/client/src/client.ts
```diff
@@ -823,6 +823,10 @@ export class SkipClient {
   }
 
   private async getAssets(chainId?: string) {
+    if (chainId) {
+      if (this.skipAssets?.[chainId]) return this.skipAssets;
+      return await this.getMainnetAndTestnetAssets(chainId);
+    }
```
toddkao, line 829: maybe something like this is a little simpler?

if (chainId && this.skipAssets?.[chainId] || !chainId && this.skipAssets) {
  return this.skipAssets;
}
return await this.getMainnetAndTestnetAssets(chainId);

toddkao: also changeset

toddkao (approved):

PR #970, opened 2025-03-28
[FRE-1549] allow connecting to any ecosystem wallet regardless of selected chain 

daps94:
- Update wallet connection flow to show all ecosystem options even when a chain is selected
-Use chainId parameter to pre-select the appropriate chain when user chooses its ecosystem
- Add automatic chain switching when reversing transaction direction
- Refactor chain switching logic for better maintainability

This change improves user experience by removing the need to change routes before connecting wallets from different ecosystems. Users can now access all wallet connection options in one place, regardless of which chain is currently selected in the swap interface.


toddkao (commented):
code looks good, only thing is I'd ask product about whether they are happy with this change in behavior to always show `ConnectedWalletModal` instead of directly showing `WalletSelectorModal` for the selected source asset 

in the linear the idea was to render two `<Connect Eco />` for the other ecosystems within `WalletSelectorModal` as two extra options at the bottom of the list, which is a slightly different UX

packages/widget/src/pages/SwapPage/SwapPage.tsx
```diff
@@ -82,6 +82,23 @@ export const SwapPage = () => {
...
+        });
+      }
+    }
+  }, [chains, connector, evmAddress, evmChainId]);
+
+  // Effect to automatically switch chain when source asset changes (including direction reversal)
+  useEffect(() => {
+    switchToEvmChainIfNeeded(sourceAsset?.chainID);
+  }, [sourceAsset?.chainID, switchToEvmChainIfNeeded]);
+
```
toddkao, line 101: nice catch 👍

packages/widget/src/modals/ConnectedWalletModal/ConnectedWalletModal.tsx
```diff
@@ -31,7 +31,12 @@ import { track } from "@amplitude/analytics-browser";
 const ITEM_HEIGHT = 60;
 const ITEM_GAP = 5;
 
-export const ConnectedWalletModal = createModal((_modalProps: ModalProps) => {
+export type ConnectedWalletModalProps = ModalProps & {
+  chainId?: string;
+};
+
+export const ConnectedWalletModal = createModal((modalProps: ConnectedWalletModalProps) => {
+  const { chainId } = modalProps;
```
toddkao, line 39: nit: you can also just use sourceAsset?.chainID instead of passing it as a prop

toddkao (approved):

PR #981, opened 2025-04-02
[FRE-1608] Add gas info for gas station transactions

ericHgorski:


packages/widget/src/pages/SwapExecutionPage/SwapExecutionPageRouteDetailed.tsx
```diff
@@ -191,6 +193,11 @@ export const SwapExecutionPageRouteDetailed = ({
           index={0}
         />
         {renderOperations}
+        {isGasStationTx && 
+        <StyledGasStationTxText>
+          Transactions from EVM to Babylon have gas provided automatically.
+        </StyledGasStationTxText>
+        }
```
codingki, line 200: As far as I know we provide gas if they have 0 BABY on Babylon. might want to change the wording "if the user don't have BABY" or we can check BABY balance first and if they don't have BABY we can show this warning

ericHgorski: pushed a fix to check balance. Its not ideal however if we've only fetched balances for EVM because source must be EVM. But better to display the message in more cases than not

packages/client/src/client.ts
```diff
@@ -86,6 +86,7 @@ import { MsgInitiateTokenDeposit } from "./codegen/opinit/ophost/v1/tx";
 
 export const SKIP_API_URL = "https://api.skip.build";
 
+export const GAS_STATION_CHAIN_IDS = ["bbn-test-5", "bbn-1", "11155111"];
```
codingki, line 89: you can't use `GAS_STATION_CHAIN_IDS`, because `11155111` was a hack to skip gas validation

codingki (approved):

PR #983, opened 2025-04-02
Filter Sei Cosmos unless Cosmos wallet explicitly connected.

codebycarson:
This PR filters out the Cosmos version of Sei unless the user explicitly connects a Cosmos wallet. Additionally it renames the assets to "Sei via EVM" and "Sei via Cosmos" if both wallets are connected and says just "Sei" otherwise.

This is a hacky approach and not dynamic, but is intended to show the functionality Sei believes is best in the Skip widget. Showing both the Cosmos and EVM versions aren't ideal as it is confusing for most Sei users who never touch Cosmos at all. This leaves the ability for Cosmos transactions to happen, but only when the user has explicitly shown they are a Cosmos user by connecting their Cosmos wallet after connecting their EVM wallet.


toddkao:
@codebycarson I've merged a PR that was based off of yours with some minor changes, hiding it behind a prop and I've let the api team know that we want to update the Sei asset names from Sei - Evm / Sei - Cosmos => Sei via Cosmos / Sei via Evm

 https://github.com/skip-mev/skip-go/pull/1040

PR #993, opened 2025-04-04
[FRE-1626, FRE-1565, FRE-1619] Fix evm disconnect, Fix useSwitchChainIfNeeded

toddkao:


packages/widget/src/hooks/useCreateEvmWallets.tsx
```diff
@@ -56,18 +57,16 @@ export const useCreateEvmWallets = () => {
 
           try {
             if (isEvmConnected && connector.id !== currentEvmConnector?.id) {
-              await currentConnector?.disconnect();
+              await disconnect(config);
```
toddkao, line 60: we need to use the disconnect function from wagmi instead of the disconnect function on the connector itself

packages/widget/src/hooks/useSwitchEvmChainIfNeeded.tsx
```diff
@@ -1,32 +1,30 @@
...
-import { useGetAccount } from "./useGetAccount";
+import { useCallback, useEffect } from "react";
 import { useAccount } from "wagmi";
 import { sourceAssetAtom } from "@/state/swapPage";
 import { useAtomValue } from "jotai";
 import { skipChainsAtom } from "@/state/skipClient";
+import { evmWalletAtom } from "@/state/wallets";
 
 export const useSwitchEvmChainIfNeeded = () => {
-  const getAccount = useGetAccount();
```
toddkao, line 9: `useGetAccount` is just really shitty and will cause issues when its included in a dependency array of a useEffect

daps94 (approved):

PR #1031, opened 2025-04-08
[FRE-1650, FRE-1522] Update History page design and reduce API calls

toddkao:
Show loading spinner if last tx in list has not settled yet

image

Show recommendedSymbol for sourceAsset, show symbol on chainName for dest asset

image

on mobile, only show recommendedSymbol 

image

added < 0.01
image

### DIFF
--- a/docs/widget/configuration.mdx
+++ b/docs/widget/configuration.mdx
@@ -237,6 +237,10 @@ Enable Amplitude analytics for the widget to improve user experience. Default: f
 Disables shadow dom, useful if there are issues with libraries not supporting shadow-dom or for enabling server side rendering
 Default: false. (shadow dom is enabled by default to avoid styling conflicts/issues)
 
+### `hideAssetsUnlessWalletTypeConnected`
+
+Filters assets based on connected wallet types (currently only supports Sei Cosmos/EVM). Added in v3.7.3. Default: false.
+
 ### `callbacks`
 
 Event handling functions.


packages/widget/src/hooks/useTxHistory.ts
```diff
@@ -42,13 +84,13 @@ export const useTxHistory = ({
       if (statusData?.lastTxStatus === "success") return "completed";
       if (isFetching && isPending) return "unconfirmed";
       if (statusData?.isSuccess) return "completed";
-      if (statusData?.isSettled && !statusData?.isSuccess) return "failed";
+      if ((statusData?.isSettled && !statusData?.isSuccess) || !chainIdFound) return "failed";
```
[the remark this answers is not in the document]
toddkao: this is to avoid calling the API if tx chainId is not found, typically this occurs if the tx history item is from another env, mainnet/testnet

packages/widget/src/hooks/useTxHistory.ts
```diff
@@ -1,35 +1,63 @@
...
+    txs?.map((tx) => tx.chainID).includes(chain.chainID),
+  );
+
+  const txsRequired = txHistoryItem?.route?.txsRequired;
+
+  const shouldFetchStatus =
+    txHistoryItem?.lastTxStatus !== "failed" &&
+    txHistoryItem?.lastTxStatus !== "success" &&
+    txs !== undefined &&
+    chainIdFound;
```
toddkao, line 36: we cant rely on `isSettled`, because it's possible that `isSettled` will never resolve, if the second tx doesn't exist / was never added to the txHistoryItem

toddkao: or actually maybe i can add sometype of limit of number of req that we do to useQuery

toddkao: with 2tx routes i think this might be problematic

toddkao: maybe i can just check txs.length and txsRequired

PR #1034, opened 2025-04-09
[FRE-1657] route preference copy change for clarity 

ericHgorski:


toddkao (approved):

.changeset/large-hornets-pretend.md
```diff
@@ -0,0 +1,5 @@
+---
+"@skip-go/widget": patch
+---
+
+copy change to route preference settings
```
toddkao, line 5: nit: maybe mention its for a tooltip copy

PR #1035, opened 2025-04-09
Use rAF timers to handle closeModal to avoid flickering

toddkao:


packages/widget/src/utils/countdownTimer.ts
```diff
@@ -14,14 +14,14 @@ export function createCountdownTimer({
     if (!start) start = timestamp;
 
     const elapsed = timestamp - start;
-    const remainingTime = Math.max(duration * 1_000 - elapsed, 0);
+    const remainingTime = Math.max(duration - elapsed, 0);
 
-    onUpdate(remainingTime / 1_000);
+    onUpdate?.(remainingTime);
```
toddkao, line 19: we used to take seconds, i had to update it to take miliseconds

packages/widget/src/components/Modal.tsx
```diff
@@ -26,15 +27,15 @@ export const Modal = ({ children, drawer, container, onOpenChange, theme }: Moda
...
-    delay(75).then(() => {
-      modal.remove();
-    });
+
+    createCountdownTimer({
+      duration: 75,
+      onComplete: () => {
+        modal.remove();
+      },
+    }).startCountdown();
```
toddkao, line 38: we can just re-use the `createCountdownTimer` i created already

codingki:
clean, lgtm

PR #1052, opened 2025-04-11
Fix connect eco row

toddkao:
if ChainType is cosmos, show addresses in this order
1. address for sourceAsset
2. address for default of ecosystem (cosmoshub-4 / provider)
3. first address found in cosmosAccounts

image

Add default chainId for ecosystem as fallback if sourceAsset?.chainId is undefined when connecting wallet


codingki (approved):
lgtm

PR #1056, opened 2025-04-11
[FRE-1646] Fix modal flickering

toddkao:
https://developer.mozilla.org/en-US/docs/Web/CSS/animation-fill-mode

Pauses animation after it completes using `animation-fill-mode: forwards`

Fixes modal flickering by pausing the animation after it completes

Remove using timer to close the modal, instead close after animation completes with `onAnimationEnd`


codingki (approved):
lgtm

PR #1058, opened 2025-04-11
[FRE-1629] add initia pubkey to accountparser in skip goclient library

codingki:


packages/client/src/registry.ts
```diff
@@ -56,7 +59,24 @@ export const accountParser: AccountParser = (acc) => {
...
+          const { key } = PubKey.decode(pubKey.value);
+          const pk = encodeSecp256k1Pubkey(key);
+          return {
+            address,
+            pubkey: pk,
+            accountNumber: Number(accountNumber),
+            sequence: Number(sequence),
+          };
+        }
+        return accountFromAny(acc);
```
toddkao, line 77: nit: can probabbly remove this return since you have it again below

toddkao (approved):

PR #1075, opened 2025-04-15
[FRE-1673]  improve button label handling if not enough to pay for fees

ericHgorski:


toddkao (commented):
fix the missing deps that the eslint caught, otherwise should be good

toddkao (approved):

PR #1076, opened 2025-04-16
History current pending tx indicator

toddkao:
image


codingki (commented):
I don't think using `overallStatus` is good. How about create a hook called `useLastTxHistory` 
inside there you get last tx history item and run `useBroadcastedTxs` with that tx history item. then only enable the hook if the tx required === 1

packages/widget/src/pages/SwapPage/SwapPageHeader.tsx
```diff
@@ -1,71 +1,81 @@
...
+    index: transactionhistory.length - 1,
+  });
 
-const CenterContainer = styled.div`
-  position: absolute;
-  left: 50%;
-  transform: translateX(-50%);
-`;
+  return null;
+});
```
toddkao, line 81: @codingki i'm tracking the latestTxHistoryItemStatus here, conditionally on `isFetchingLastTransactionStatus` which current is 

```ts
  return (
    overallStatus === "pending" ||
    (lastTxHistoryItem?.isSettled !== true && lastTxHistoryItem?.route?.txsRequired === 1)
  );
```
  
  if we remove `overallStatus` it would be what i had before i added overallStatus as a way of make the spinner show up essentially instantly (before the lastTxHistoryItem is updated, because there's currently a delay)
  ```ts
  lastTxHistoryItem?.isSettled !== true && lastTxHistoryItem?.route?.txsRequired === 1
  ```

toddkao: there's a couple unintuitive things here i'm doing to avoid unnecessary re-renders as well as allowing a hook to be conditionally called

codingki (approved):

PR #1088, opened 2025-04-23
[FRE-1679] Refactor fee calculation logic and display fees in settings bar

ericHgorski:


packages/widget/src/pages/SwapPage/SwapPageFooter.tsx
```diff
@@ -144,6 +156,9 @@ export const SwapPageFooterItems = ({
                 </Row>
               </>
             )}
+            <Row gap={8} align="flex-end">
+              {totalFees?.formattedUsdAmount}
```
codingki, line 160: add `totalFees?.formattedUsdAmount` to the hook dependancy

ericHgorski:
> also i forget did we decide to show these fees only in the footer? i thought we wanted them to show only next to the % difference image

We settled it during an in-office convo! But still looking to iterate on this.

toddkao (approved):

PR #1113, opened 2025-04-30
Fix clearing validating when returning from user rejected request, Fix no longer showing user rejected request unless currently on swap execution page

toddkao:
Fix clearing validating when returning from user rejected request, Fix no longer showing user rejected request unless currently on swap execution page


ericHgorski (approved):
LGTM

PR #1114, opened 2025-04-30
Add onSourceAndDestinationSwapped callback prop

toddkao:


packages/widget/src/state/swapPage.ts
```diff
@@ -165,7 +166,16 @@ export const invertSwapAtom = atom(null, (get, set) => {
...
+
     set(isInvertingSwapAtom, false);
+    callbacks?.onSwappedSourceAndDestinationAssets?.({
+      srcChainId: sourceAsset?.chainID,
+      srcAssetDenom: sourceAsset?.denom,
+      destChainId: destinationAsset?.chainID,
+      destAssetDenom: destinationAsset?.denom,
+      amountIn: sourceAsset?.amount,
+      amountOut: destinationAsset?.amount,
+    });
```
codingki: this change is sending a previous value to the callback, not sure if it's itentional

```suggestion
    callbacks?.onSwappedSourceAndDestinationAssets?.({
      srcChainId: destinationAsset?.chainID,,
      srcAssetDenom: destinationAsset?.denom,
      destChainId: sourceAsset?.chainID,
      destAssetDenom: sourceAsset?.denom,
      amountIn: destinationAsset?.amount,
      amountOut: sourceAsset?.amount,
    });
```

toddkao: ah you're right, damn i assumed it would have swapped by the time i called the callback but i guess not... this feels weird to me, maybe it'll be safer to clone source/dest assets first

codingki (approved):

PR #1115, opened 2025-04-30
Add screenshots to failed e2e test cases

toddkao:
added a new util `saveScreenshots` that will handle both updating the expected images as well as saving a new screenshot only when the `expectFn` throw an error

added a script `combine-images.mjs` that will merge expected image, diff as well as a new screenshot (when an error occurs) and upload it in the artifact for `lumberjack-bot` to comment in a PR

here is an example of what would be shown by the bot if there's an error within one of the tests:

image


codingki:
if the amount out changes in the next test, is it going to fail the test?

PR #1116, opened 2025-05-01
update injective deps

ericHgorski:
update injective deps and reduce bundle size


toddkao (approved):

bangjelkoski:
Can we republish the package version?

PR #1120, opened 2025-05-02
[FRE-1336] Update fee-info documentation

daps94:
Update fee-info documentation to clarify included fees and current lmitations, specifying supported bridge fees and noting future expansions.


ericHgorski (approved):
LGTM

PR #1139, opened 2025-05-08
[FRE-1758] Amplitude metrics cleanup

ericHgorski:


packages/widget/src/widget/initAmplitude.ts
```diff
@@ -13,7 +13,9 @@ export const initAmplitude = () => {
   init("14616a575f32087cf0403ab8f3ea3ce0", {
     autocapture: true,
     appVersion: version,
+    // serverUrl: "https://go.skip.build/api/amplitude"
```
toddkao, line 16: nit: should we remove this if we're not using it? typically we dont keep commented out code around

toddkao (approved):

packages/widget/src/widget/initAmplitude.ts
ericHgorski: Im working on a separate PR on the app side to get this setup. So itll be commented out shortly!

PR #1144, opened 2025-05-12
Add executeRoute tab to docs

toddkao:


docs/client/executeRoute.mdx
```diff
@@ -0,0 +1,81 @@
+---
+title: 'executeRoute'
```
codingki, line 2: I think "Executing a route" is a better title

codingki (approved):

toddkao:
@codingki does that latest change work? i thought it was based on the file name? we'd need to change the filename too or it's pointing to an invalid file

PR #1171, opened 2025-05-14
fix client build

toddkao:
Update all imports to follow esm convention
Update codegen to patch protobuf/minimal.js imports


packages/client/scripts/codegen.cjs
```diff
@@ -236,4 +237,28 @@ const mergeArrays = (arr1, arr2) => {
...
+        'import $1 from "protobufjs/minimal.js";'
+      );
+
+      if (updated !== content) {
+        await fs.writeFile(fullPath, updated, "utf-8");
+      }
+    }
+  }
+}
+
```
toddkao, line 263: this function manually patches protobufjsImports to follow esm import convention, after telescope generates them

codingki (approved):

PR #1189, opened 2025-05-16
Temporarily hide close animation

toddkao:


packages/widget/src/pages/SwapPage/SwapPage.tsx
codingki, line 1: in your testing, is it only in asset modal? how about the connect wallet modal and other?

toddkao: i wasn't able to repro the issue with other modals, just this one

ericHgorski (approved):

PR #1207, opened 2025-05-20
[API-4122] Update release notes for API, client library, and widget; 

daps94:
Update release notes for API, client library, and widget;

### DIFF
--- a/docs/release-notes/api.mdx
+++ b/docs/release-notes/api.mdx
@@ -3,6 +3,18 @@ title: "REST API"
 description: "Selected updates and improvements to the Skip Go REST API."
 ---
 
+## May 2025
+- **Improved Cross-Chain Composability:** Execute EVM swaps immediately following a Eureka cross-chain transfer, enabling more sophisticated multi-step routes.
+- **Enhanced Fee Transparency:** Fees associated with Eureka operations are now clearly displayed within the estimated transaction fees for better cost predictability.
+- **LayerZero and Stargate Integration:** Expanded cross-chain capabilities with the integration of LayerZero and Stargate protocols.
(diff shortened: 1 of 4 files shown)


docs/release-notes/client.mdx
```diff
@@ -1,12 +1,26 @@
...
+description: "Selected updates and improvements to the `@skip-go/client` [npm package](https://www.npmjs.com/package/@skip-go/client)."
 ---
 
+## v1.1.0
+
+### Enhancements
+- **Transaction Timestamps:** The `ChainTransaction` type now includes an `on_chain_ts` field, providing a timestamp for when the transaction was confirmed on-chain. This can be useful for tracking and display purposes.
+- **Exported JSON Types:** Relevant JSON types used by the client are now exported, improving the developer experience for those working directly with these underlying types.
+
+## v1.0.0
```
womensrights, line 12: should we indicate that we don't recommend moving to v1 and straight to v1.1.0 somewhere?

docs/release-notes/api.mdx
```diff
@@ -3,6 +3,18 @@ title: "REST API"
 description: "Selected updates and improvements to the Skip Go REST API."
 ---
 
+## May 2025
+- **Improved Cross-Chain Composability:** Execute EVM swaps immediately following a Eureka cross-chain transfer, enabling more sophisticated multi-step routes.
+- **Enhanced Fee Transparency:** Fees associated with Eureka operations are now clearly displayed within the estimated transaction fees for better cost predictability.
```
womensrights, line 8: should we also add this to the widget? New fees display for all routes or something like that - I think its quite a big feature

daps94: updated the widget release notes

docs/release-notes/client.mdx
daps94: added a call out for using the latest version

PR #1221, opened 2025-05-21
FRE-1764 track warnings not as errors separate warnings page

codingki:
`error` renamed to `blockingPage`


ericHgorski:
> `error` renamed to `blockingPage`

Do we prefer `blockingPageAtom` to just `blockingAtom` ?

toddkao:
i'm thinking maybe we should have two folders at the top level:

`ExpectedErrors`
`UnexpectedErrors`

or something like that

and the existing `ErrorPageUnexpected` , `TransactionFailed` ,`TransactionReverted` (whichever ones are classified as errors) should be in the `UnexpectedErrors` and the others in the `ExpectedErrors`

should be easy to also split which ones are tracked as expected or unexpected

toddkao:
i think we should just keep it as `ErrorPage` at the top level

also typo on the `ExpectedErrorpage` folder

packages/widget/src/state/route.ts
```diff
@@ -104,20 +104,22 @@ export const routeConfigAtom = atom<WidgetRouteConfig>({
...
-  const swapSettings = get(swapSettingsAtom)
+  const swapSettings = get(swapSettingsAtom);
 
   const queryEnabled =
     params !== undefined &&
     (Number(params.amountIn) > 0 || Number(params.amountOut) > 0) &&
     !isInvertingSwap &&
     currentPage === Routes.SwapPage &&
-    error === undefined;
+    blockingPage === undefined;
```
toddkao, line 122: nit: some `blockingPage` still here

toddkao (approved):

PR #1225, opened 2025-05-21
Add transactionHistory as soon as transaction is broadcasted

toddkao:


packages/widget/src/state/swapExecutionPage.ts
```diff
@@ -165,6 +165,14 @@ export const setSwapExecutionStateAtom = atom(null, (get, set) => {
...
         transactionHistoryIndex,
       );
+
+      set(setTransactionHistoryAtom, transactionHistoryIndex, {
+        route: route as RouteResponse,
+        transactionDetails: [txInfo],
+        timestamp: Date.now(),
+        status: "unconfirmed",
+      });
+
```
codingki, line 175: This is rendundant, because we already did it in line 163-167.

codingki:
unless we want to remove this line https://github.com/skip-mev/skip-go/blob/055bd1f35a1f767fef21830d1c71cf1152bab881/packages/widget/src/state/swapExecutionPage.ts#L318-L323

packages/widget/src/state/swapExecutionPage.ts
toddkao: you're right, i misread that function and thought it didn't take the txInfo and put it in transactionDetails, but it actually does

PR #1250, opened 2025-05-23
Fix gas balance check

NotJeremyLiu:
## Summary
- fix early return logic in validateGasBalances
- add changeset

## Testing
- `npm test --silent` *(fails: Error when performing the request to https://repo.yarnpkg.com/...)*

### DIFF
--- a/.changeset/fix-gas-validation.md
+++ b/.changeset/fix-gas-validation.md
@@ -1,0 +1,7 @@
+---
+"@skip-go/client": patch
+---
+Fix logic that detected absence of Cosmos and SVM transactions in `validateGasBalances`.
+Previously the check always failed, causing gas balance validation to run even when only
+EVM transactions were present. This patch ensures the function exits early when there are
+no Cosmos or SVM transactions to validate.
--- a/.changeset/fix-gas-validation.md
+++ b/.changeset/fix-gas-validation.md
@@ -1,7 +1,4 @@
 ---
 "@skip-go/client": patch
 ---
-Fix logic that detected absence of Cosmos and SVM transactions in `validateGasBalances`.
-Previously the check always failed, causing gas balance validation to run even when only
-EVM transactions were present. This patch ensures the function exits early when there are
-no Cosmos or SVM transactions to validate.
+validate all txs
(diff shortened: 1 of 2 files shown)


packages/client/src/private-functions/validateGasBalances.ts
```diff
@@ -29,10 +29,10 @@ export const validateGasBalances = async ({
   enabledChainIds,
   useUnlimitedApproval,
 }: ValidateGasBalancesProps) => {
-  // cosmos or svm tx in txs
+  // return early if there are no cosmos or svm transactions to validate
   if (
-    txs.every((tx) => "cosmosTx" in tx === undefined) ||
-    txs.every((tx) => "svmTx" in tx === undefined)
+    txs.every((tx) => !("cosmosTx" in tx)) &&
+    txs.every((tx) => !("svmTx" in tx))
```
codingki, line 35:
```suggestion
    txs.every((tx) => !("cosmosTx" in tx)) ||
    txs.every((tx) => !("svmTx" in tx))
```

packages/client/src/private-functions/validateGasBalances.ts
```diff
@@ -29,10 +29,10 @@ export const validateGasBalances = async ({
...
   ) {
     return;
   }
```
codingki, line 38: we can remove this if block. because we are now validating evm tx too

.changeset/fix-gas-validation.md
```diff
@@ -0,0 +1,7 @@
+---
+"@skip-go/client": patch
+---
+Fix logic that detected absence of Cosmos and SVM transactions in `validateGasBalances`.
+Previously the check always failed, causing gas balance validation to run even when only
+EVM transactions were present. This patch ensures the function exits early when there are
+no Cosmos or SVM transactions to validate.
```
codingki, line 7:
```suggestion
validate all txs
```

codingki (changes requested):
commit suggestion and should be good to go

NotJeremyLiu:
updated for suggestions

codingki (approved):

PR #1280, opened 2025-05-28
 Only pass specific fields to msgs api request

toddkao:


packages/client/src/api/postMessages.ts
```diff
@@ -1,11 +1,30 @@
...
+    chainIdsToAffiliates: request.chainIdsToAffiliates,
+    enableGasWarnings: request.enableGasWarnings,
+  };
+
+  return api({
+    methodName: "msgs",
+    method: "post",
+    path: "/v2/fungible/msgs",
+  })(messagesRequest);
+};
```
toddkao, line 27: honestly this feels like regressing back to the old way of doing things, and I realize it's just one exception but it's still annoying

toddkao: I feel like the API should just ignore extra fields if they are passed

toddkao: or honestly just telling integrators they can't just pass the entire `route` response to `msgs` (cuz you can't)

codingki (approved):

PR #1285, opened 2025-05-29
Add 15s timeout before expecting transfer asset release in case of failure

toddkao:


packages/widget/src/pages/SwapExecutionPage/useSwapExecutionState.ts
```diff
@@ -38,7 +38,7 @@ export function useSwapExecutionState({
       return SwapExecutionState.confirmed;
     }
 
-    if (overallStatus === "pending") {
+    if (overallStatus === "pending" || overallStatus === "failed") {
```
toddkao, line 41: this is so the edge case of us waiting the extra 5 seconds we continue to show "pending" in the `MainButton`, currently we show `Validating` if there's an error

PR #1286, opened 2025-05-29
[FRE-1337] Implement submit endpoint for Cosmos transactions and update transact…

ericHgorski:
…ion broadcasting


packages/client/src/api/postSubmit.ts
```diff
@@ -0,0 +1,11 @@
...
+import type { ApiRequest, ApiResponse } from "../utils/generateApi";
+
+export const submit = api({
+  methodName: "submit",
+  method: "post",
+  path: "/v2/tx/submit",
+});
+
+export type MessagesRequest = ApiRequest<"submit">;
+export type MessagesResponse = ApiResponse<"submit">;
```
toddkao, line 11: these should be SubmitRequest / SubmitResponse

packages/client/src/private-functions/cosmos/executeCosmosTransaction.ts
```diff
@@ -109,11 +110,15 @@ export const executeCosmosTransaction = async ({
   });
 
   const txBytes = TxRaw.encode(rawTx).finish();
+  const rawTxBase64 = Buffer.from(txBytes).toString("base64");
 
-  const txResponse = await stargateClient.broadcastTx(txBytes);
+  const txResponse = await submit({
+    chainId,
+    tx: rawTxBase64,
+  })
```
toddkao, line 118: i'm curious if we use submit endpoint how long does it take for the keplr tx window to popup compared to before where we pre-loaded the `stargateClient`

packages/client/src/api/postSubmit.ts
ericHgorski: Fixed!

PR #1289, opened 2025-05-30
Fix zero seconds handling

CryptoAssassin1:
## Summary
- handle `0` seconds in `convertSecondsToMinutesOrHours`
- add test covering zero seconds case
- add changeset entry

## Testing
- `npm test` *(fails: EHOSTUNREACH)*

### DIFF
--- a/packages/widget/src/utils/number.ts
+++ b/packages/widget/src/utils/number.ts
@@ -30,7 +30,7 @@ export function calculatePercentageChange(
 }
 
 export const convertSecondsToMinutesOrHours = (seconds?: number) => {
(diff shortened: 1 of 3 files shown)


packages/widget/src/utils/number.ts
```diff
@@ -30,7 +30,7 @@ export function calculatePercentageChange(
 }
 
 export const convertSecondsToMinutesOrHours = (seconds?: number) => {
-  if (!seconds) {
+  if (seconds === undefined) {
```
codingki, line 33: it making it only return when the value is undefined

PR #1291, opened 2025-05-30
Fix async wallet address loop

CryptoAssassin1:
## Summary
- fix async loop in `useAutoSetAddress`
- add changeset for widget

## Testing
- `yarn test` *(fails: EHOSTUNREACH)*


codingki:
this is basically the same thing

PR #1301, opened 2025-06-02
add storing extra chainIdsPerWallet in localStorage and auto-clearing it and retrying if there's an error when attempting to connect wallet

toddkao:


ericHgorski (approved):
LGTM

PR #1313, opened 2025-06-04
[FRE-1883] Update widget to allow getCosmosSigner/getEvmSigner to fallback to default if no signer is found for given chainId

toddkao:


codingki (approved):
good call, lgtm

PR #1317, opened 2025-06-04
Fix infinite spinner in the history icon

toddkao:
Fix infinite spinner for the history icon if the last transaction was a 2tx route but the second tx has not been signed


ericHgorski (approved):
LGTM

PR #1356, opened 2025-06-10
[FRE-1911] Use timestamp as unique id for transactionHistoryItem

toddkao:
Use timestamp as unique id for `transactionHistoryItem`

Added `lastTransactionInTimeAtom` to correctly grab the lastTransaction based on timestamp instead of simply the last txHistoryItem in the array

### DIFF
--- a/.changeset/fast-years-notice.md
+++ b/.changeset/fast-years-notice.md
@@ -1,0 +1,5 @@
+---
+"@skip-go/widget": patch
+---
+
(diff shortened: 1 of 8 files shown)


codingki:
got this error when opening the history page 
image

toddkao:
> got this error when opening the history page image

damn it, i guess the timestamp must be undefined for one of ur history items??

codingki (approved):

PR #1362, opened 2025-06-11
Add useCroppedImage, Added shimmer effect to Skeleton

toddkao:
Add useCroppedImage
Added shimmer effect to Skeleton

image


packages/widget/src/utils/style.ts
```diff
@@ -0,0 +1,6 @@
+export const toPxValue = (value?: string | number): string => {
+  if (value === undefined || value === null) {
```
solophbs, line 2: if (!value) {

I think this will work as well.

codingki (approved):

PR #1367, opened 2025-06-12
add throwOnError to pollingApi, return failed txsStatus if error is caught when trying to get transactionStatus

toddkao:


packages/widget/src/pages/SwapExecutionPage/useBroadcastedTxs.ts
```diff
@@ -35,46 +35,60 @@ export const useBroadcastedTxsStatus = ({
...
+        const resData: TxsStatus = {
+          isSuccess: false,
+          isSettled: true,
+          lastTxStatus: "failed",
+          transferEvents: [],
+          transferAssetRelease: undefined,
+        };
+        setPrevData(resData);
+        return resData;
+      }
```
toddkao, line 91: this is basically the entire change, just wrapping this with a try/catch and returning this if an error is found

codingki (approved):

PR #1369, opened 2025-06-12
store history in indexedDB instead of localStorage

toddkao:


codingki:
We are using the same structure with previous implementation
image

is it going to be the same if the value is too many?
I thought we are going to store timestamp as key and value is the data

toddkao:
> Oh we use indexedDB because it having an error if the array too big right?

yeah it's just so we have a larger storage size

codingki (approved):
I'll approve this with a note that I think storing it as key value is better maybe we can do it next time.

nit: fix eslint error and add changeset

PR #1373, opened 2025-06-16
[FRE-1890] solana wallet provider

codingki:


packages/widget/src/state/swapExecutionPage.ts
```diff
@@ -450,11 +450,11 @@ export const skipSubmitSwapExecutionAtom = atomWithMutation((get) => {
...
-            const walletName = wallets.svm?.walletName;
-            if (!walletName) throw new Error("getSvmSigner error: no svm wallet");
-            const solanaWallet = solanaWallets.find((w) => w.name === walletName);
-            if (!solanaWallet) throw new Error("getSvmSigner error: wallet not found");
-            return solanaWallet;
+            const adapter = await getSvmSigner();
+            if (!adapter) {
+              throw new Error("getSvmSigner error: no SVM wallet");
+            }
+            return adapter;
```
toddkao, line 457: @codingki why can't we handle this in here instead of passing it through as a prop?

toddkao (approved):
I want to refactor this so we don't have to pass `getSvmSigner` in as a prop but that can be in a future PR

PR #1411, opened 2025-06-23
Remove redundant submit function (use submitTransaction instead), remove track call for non-evm txs, increase default maxRetries to 10

toddkao:
- Remove redundant submit function
- remove track call for non-evm txs
- increase default maxRetries to 10
- small waitForTransactionRefactor


packages/client/src/private-functions/executeTransactions.ts
```diff
@@ -179,17 +179,26 @@ export const executeTransactions = async (
...
     await onTransactionBroadcast?.({ ...txResult });
 
+    if (txResult.explorerLink) {
+      options.onTransactionTracked?.({
+        txHash: txResult.txHash,
+        chainId: txResult.chainId,
+        explorerLink: txResult.explorerLink ?? "",
+      });
+    }
+
```
toddkao, line 197: if `txResult.explorerLink` exists, that means we used the `/submit` endpoint, so we can treat it as tracked

codingki (approved):

PR #1432, opened 2025-06-30
Kiki/fre 1980 UI stuck on confirming in wallet with timeout error

codingki:


packages/client/src/private-functions/cosmos/signCosmosMessageDirect.ts
```diff
@@ -23,22 +21,6 @@ export const signCosmosMessageDirect = async (
...
-    });
-  }
-
-  if (chainId.includes("injective")) {
-    return signCosmosMessageDirectInjective(signerAddress, signer, cosmosMsgs, fee, {
-      accountNumber,
-      sequence,
-      chainId: signerChainId,
-    });
-  }
```
toddkao, line 40: should we remove these functions? I think we're only using them here

PR #1441, opened 2025-07-02
[FRE-1996] dont select bridges and venues by default

ericHgorski:
Sending every swap venue and bridge breaks the API for some reason

### DIFF
--- a/examples/nextjs/src/app/page.tsx
+++ b/examples/nextjs/src/app/page.tsx
@@ -161,8 +161,7 @@ export default function Home() {
                     id={venue.name}
                     type="checkbox"
                     checked={
-                      swapVenues?.map((v) => v.name).includes(venue.name) ??
-                      true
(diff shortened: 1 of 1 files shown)


examples/nextjs/src/app/page.tsx
```diff
@@ -102,17 +102,6 @@ export default function Home() {
...
-    if (swapVenuesQuery.data) {
-      setSwapVenues(swapVenuesQuery.data);
-    }
-  }, [swapVenuesQuery.data]);
-  useEffect(() => {
-    if (bridgesQuery.data) {
-      setBridges(bridgesQuery.data);
-    }
-  }, [bridgesQuery.data]);
-
```
toddkao, line 115: i think we need to still set the value for swapVenues and bridges, but default them to false

toddkao: otherwise right now none of the checkboxes do anything

PR #1454, opened 2025-07-08
feat: single tx with no wait mode

AustinWoetzel:
- Adds support for a mode that is for signing + broadcast of single tx only and does not monitor the final state of the transaction.

Why is this useful? 

The skip client executeTransactions() function has built-in state management for monitoring a tx end-to-end and this is a very convenient feature, however this limits flexibility for users that may want to implement any sort of custom state management. It is helpful to be able to separate tx execution from tracking/waiting for the tx to be confirmed. This new mode will allow this separation for the most common case of a single tx.

FUTURE WORK: This feature could be expanded to allow for some amount of separation/isolation to support multi-tx state management but that would be a more significant PR as there is not really a simple way to create reusable functions to accomplish this.

### DIFF
--- a/packages/client/src/private-functions/executeTransactions.ts
+++ b/packages/client/src/private-functions/executeTransactions.ts
@@ -30,11 +30,18 @@ export const executeTransactions = async (
     onValidateGasBalance,
     trackTxPollingOptions,
     batchSignTxs = true,
+    singleTxNoWaitMode = false,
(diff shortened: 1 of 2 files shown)


AustinWoetzel:
Looks like I am too far behind the staging branch and recent change impact this design. I will re-open this PR once that is resolved.

PR #1476, opened 2025-07-14
Add tracking in amplitude for the destination wallet source

toddkao:


packages/widget/src/pages/SwapExecutionPage/SwapExecutionButton.tsx
```diff
@@ -129,6 +132,16 @@ export const SwapExecutionButton: React.FC<SwapExecutionButtonProps> = ({
           });
           return;
         }
+
+        const destinationWalletSource =
+          route?.requiredChainAddresses?.length &&
+          chainAddresses[route?.requiredChainAddresses?.length - 1]?.source;
+
+        if (destinationWalletSource) {
+          track("destination wallet source", {
```
codingki, line 141: I think it's better to add the value in `swap execution page: confirm button - clicked` event

PR #1480, opened 2025-07-15
Revert changes in #1460, fix hydration error by avoiding nested buttons

toddkao:


packages/widget/src/modals/ConnectedWalletModal/ConnectEcoRow.tsx
```diff
@@ -198,14 +197,14 @@ export const ConnectEcoRow = ({
             </Row>
           </Row>
         ) : (
-          <TextButton>
+          <Text>
```
toddkao, line 200: This is the real fix to the hydration error

PR #1506, opened 2025-07-28
fix loading state

codingki:


packages/client/src/public-functions/subscribeToRouteStatus.ts
```diff
@@ -262,24 +262,34 @@ export const executeAndSubscribeToRouteStatus = async ({
...
+        if (
+          updatedRouteDetails?.relatedRoutes &&
+          updatedRouteDetails?.relatedRoutes?.length > 0
+        ) {
+          if (allRelatedRoutesFinal) {
+            break;
+          }
+        } else {
+          break;
+        }
```
toddkao, line 288: something like this would be easier to read
if (
  isFinalState(transaction) &&
  (!updatedRouteDetails?.relatedRoutes?.length || allRelatedRoutesFinal)
) {
  break;
}

toddkao (approved):

PR #1513, opened 2025-07-29
gor amounts fix

codingki:


packages/widget/src/state/gasOnReceive.ts
```diff
@@ -116,14 +116,14 @@ export const isSomeDestinationFeeBalanceAvailableAtom = atomWithQuery((get) => {
...
           const balanceAmount =
             balanceResponse?.chains?.[destination.chainId]?.denoms?.[denom]?.amount;
           const isMoreThanAmountIn =
             !!balanceAmount &&
-            !!amountIn &&
-            BigNumber(balanceAmount).isGreaterThanOrEqualTo(amountIn);
-          return amountIn ? isMoreThanAmountIn : balanceAmount && balanceAmount !== "0";
+            !!amountOut &&
+            BigNumber(balanceAmount).isGreaterThanOrEqualTo(amountOut);
+          return amountOut ? isMoreThanAmountIn : balanceAmount && balanceAmount !== "0";
```
toddkao, line 126: i dont understand why we are changing to check balance is more than `amountOut` instead of `amountIn`? it seemed fine before?

PR #1524, opened 2025-08-01
Hot fix: switch component & numbers fix

codingki:


packages/widget/src/components/Switch.tsx
```diff
@@ -31,18 +31,14 @@ type SwitchProps = {
...
-    setIsChecked(newVal);
-    onChange?.(newVal);
+    onChange?.(!checked);
   };
 
   return (
-    <SwitchWrapper checked={isChecked} onClick={toggle}>
-      <SwitchThumb checked={isChecked} />
+    <SwitchWrapper checked={!!checked} onClick={toggle}>
+      <SwitchThumb checked={!!checked} />
```
toddkao, line 41: @codingki i dont think the !!checked is necessary this should be fine?

```
export const Switch = ({ checked = false, onChange }: SwitchProps) => {
  const toggle = () => {
    onChange?.(!checked);
  };

  return (
    <SwitchWrapper checked={checked} onClick={toggle}>
      <SwitchThumb checked={checked} />
    </SwitchWrapper>
  );
};
```

codingki: yeah that works

PR #1539, opened 2025-08-05
Allow calling api functions without first configuring api/client options

toddkao:
- [x] Allow calling api functions without first configuring api/client options
- [x] Throws an error message if you attempt to call `setApiOptions` or `setClientOptions` after calling one of the api functions (if you haven't called one of these functions before)
- [x] Allow bypassing this if you pass `allowOptionsUpdateAfterApiCall` to `setApiOptions` or `setClientOptions`

image

### DIFF
--- a/.changeset/fast-plums-worry.md
+++ b/.changeset/fast-plums-worry.md
@@ -1,0 +1,5 @@
+---
+"@skip-go/client": patch
+---
+
+Allow calling api functions without first configuring api/client options
--- a/examples/client/index.ts
+++ b/examples/client/index.ts
@@ -1,7 +1,6 @@
 import { route, setApiOptions, getRecommendedGasPrice } from "@skip-go/client";
 
 const getRoute = async () => {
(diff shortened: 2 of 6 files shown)


toddkao:
> > Throws an error message if you attempt to call setApiOptions or setClientOptions after calling one of the api functions (if you haven't called one of these functions before)
> 
> Why we throw error for this? shouldn't we mutate the options if it happen?

this is to make it obvious if the race condition happens where someone is trying to call `setApiOptions` or `setClientOptions` but they called it at the same time or slightly after calling one of the api functions, which would result in calling the api with the default `apiUrl` even though they didn't want to. if you call `setApiOptions` / `setClientOptions` first before calling the api and then later call it again it won't throw an error though because `ApiState.initialized = true` after the initial call.

It's all just a way to avoid the situation where the initial call from one of the api functions is calling the `defaultUrl` unintentionally

codingki (approved):

PR #1571, opened 2025-08-15
feat(explorer): add view raw data

codingki:
CleanShot 2025-08-15 at 21 31 27


apps/explorer/src/components/modals/ViewRawData.tsx
```diff
@@ -0,0 +1,93 @@
...
+    const [isCopied, setIsCopied] = useState(false);
+
+    useEffect(() => {
+      if (!isCopied) return;
+      const timer = setTimeout(() => {
+        setIsCopied(false);
+      }, 1000);
+
+      return () => clearTimeout(timer);
+    }, [isCopied]);
```
toddkao, line 26: we can reuse the `useCopyAddress` hook for this, although maybe we should rename the hook to not be about addresses

PR #1654, opened 2025-09-10
Remove RoutePreferenceSelector if goFast: false is passed

toddkao:
emove RoutePreferenceSelector if goFast: false is passed
set routePreference to "cheapest" automatically if goFast:false is passed
image


ericHgorski (approved):
LGTM

PR #1679, opened 2025-09-24
Add chain integration request form

Cordtus:
## Add Chain Integration Request Form

  Adds a dedicated form for chains to request Skip Go API integration, replacing Discord as the
  primary path for these requests.

  ### Changes
  - New integration request form page with two paths: pre-launch chains (project info only) and post-launch chains
  (requires Chain Registry URL and endpoints)
  - Validates Chain Registry URLs via GitHub API
  - Submits to Formspree → Linear (via email) for automatic issue creation
  - Added to docs navigation under Support Requirements
  -Comms preference selection and other helpful bits of info

### DIFF
--- a/docs/snippets/SkipAPIChainForm.jsx
+++ b/docs/snippets/SkipAPIChainForm.jsx
@@ -283,8 +283,8 @@ ${useTelegram && telegramHandle ? `- Telegram: ${telegramHandle}` : ''}
       <div style={{
         marginBottom: '24px',
         padding: '12px',
-        backgroundColor: 'var(--background-secondary, #f8f9fa)',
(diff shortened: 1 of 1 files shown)


docs/docs.json
```diff
@@ -81,7 +81,8 @@
             "pages": [
               "support-requirements/chain-support-requirements",
               "support-requirements/token-support-requirements",
-              "support-requirements/swap-venue-requirements"
+              "support-requirements/swap-venue-requirements",
+              "support-requirements/request-form"
```
NotJeremyLiu, line 85: can we name this to chain-integration-request

PR #1768, opened 2026-07-13
feat: support per-environment Skip explorer deployments

mintae-clk:
## Summary

Makes the Skip explorer deployable per environment (dev / testnet) instead of
being pinned to the production instance. Three related changes:

**Explorer app — API URL and testnet mode from env vars**
`apps/explorer` hardcoded `https://api.skip.build`. It now reads
`NEXT_PUBLIC_API_URL` (falling back to the default client config) and
`NEXT_PUBLIC_IS_TESTNET`. The `is_testnet` query param lost its `false`
default, so a URL param still wins when present, but the env var takes over
when it is absent. The reindex call (`/v2/tx/retry_track`) now uses the same
configured API URL.

**Widget — new optional `skipExplorerUrl` prop**
Transaction history links were hardcoded to `https://explorer.skip.build`.
The widget now accepts an optional `skipExplorerUrl` prop, wired through a
Jotai atom into `createSkipExplorerLink`. Defaults to
`https://explorer.skip.build`, so existing integrations are unchanged.
Exposed on the web component as well. Changeset included (minor).

**CI — sync workflow**
`main` now opens sync PRs to `dev` and `testnet-dev` in addition to `staging`,
via a build matrix with `fail-fast: false`. Renamed `sync-staging.yml` to
`sync-branches.yml` since it no longer targets staging alone.

### DIFF
--- a/.github/workflows/sync-branches.yml
+++ b/.github/workflows/sync-branches.yml
@@ -1,4 +1,4 @@
-name: Sync (Staging)
+name: Syncing branches
 on:
   push:
     branches:
@@ -7,7 +7,11 @@ on:
 jobs:
   sync-branches:
     runs-on: ubuntu-latest
-    name: Syncing branches
+    name: Syncing branches (${{ matrix.to_branch }})
+    strategy:
+      fail-fast: false
+      matrix:
+        to_branch: [staging, dev, testnet-dev]
     steps:
       - name: Checkout
         uses: actions/checkout@v2
@@ -21,5 +25,5 @@ jobs:
         with:
           GITHUB_TOKEN: ${{secrets.PERSONAL_TOKEN}}
           FROM_BRANCH: 'main'
-          TO_BRANCH: 'staging'
+          TO_BRANCH: ${{ matrix.to_branch }}
           CONTENT_COMPARISON: true
--- a/.github/workflows/sync-branches.yml
+++ b/.github/workflows/sync-branches.yml
@@ -11,7 +11,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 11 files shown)


apps/explorer/src/app/page.tsx
```diff
@@ -62,10 +64,8 @@ export default function Home() {
     parseAsArrayOf(parseAsString, ",")
   );
 
-  const [isTestnet] = useQueryState(
-    "is_testnet",
-    parseAsBoolean.withDefault(false)
-  );
+  // No default: null means "not in the URL", so NEXT_PUBLIC_IS_TESTNET can take over.
+  const [isTestnet] = useQueryState("is_testnet", parseAsBoolean);
```
JunyupHong: I think it would be good to modify it to use `is_testnet` params only instead of adding environment variables.
It looks like it can be controlled by the `is_testnet` property in skip-go-app.

.github/workflows/sync-branches.yml
```diff
@@ -7,7 +7,11 @@ on:
 jobs:
   sync-branches:
     runs-on: ubuntu-latest
-    name: Syncing branches
+    name: Syncing branches (${{ matrix.to_branch }})
+    strategy:
+      fail-fast: false
+      matrix:
+        to_branch: [staging, dev, testnet-dev]
```
JunyupHong: Based on the comments below (`apps/explorer/src/app/page.tsx`), the testnet-related branch can be removed

JunyupHong (changes requested):
Work in the direction of using the existing `is_testnet` design,
Depending on the environment, it would be good to control only `SKIP_API_URL`.

JunyupHong (approved):

PR #1778, opened 2026-07-30
fix(explorer): resolve false "Try again" on intermediate transfer cards

mintae-clk:
## Problem
Completed multi-tx routes could show a false `Try again on Skip.go →` on an
intermediate card. Most visible on CCTP v2 → Injective: the release lands on
Injective EVM (`1439`) while the next IBC leg continues from `injective-888` —
same network, different chain-id, so it looked like a dead-end.

## Root cause
The release was attached by matching the first tx's release `chainId` to a card.
`1439 !== injective-888`, so the code missed that funds continued and pinned
"Try again" to the `1439` card.

## Fix
Use the `released` flag sequence instead of chain-id matching: find the last tx
that released while the next didn't (= where funds actually stopped), and attach
the release to that tx's destination card. `assetMatches` (denom+chainId) is kept
for same-chain swaps. On completion the release lands on the Destination card, so
"Try again" is suppressed.

### DIFF
--- a/apps/explorer/src/app/page.tsx
+++ b/apps/explorer/src/app/page.tsx
@@ -295,22 +295,10 @@ export default function Home() {
               if (tx.txHash && !trackedTxHashes.current.includes(tx.txHash)) {
                 trackedTxHashes.current.push(tx.txHash);
                 if (notFound) {
-                  try {
-                    await trackTransaction({
-                      txHash: tx.txHash,
-                      chainId: tx.chainId,
-                      maxRetries: 1,
-                    });
-                  } catch (trackError) {
-                    if (index !== 0 && destAsset) {
-                      setDestinationNodeFailed(true);
-                    }
-                    setErrorDetails({
-                      errorMessage: ErrorMessages.TRANSACTION_NOT_FOUND,
-                      error: trackError as ErrorWithCodeAndDetails,
-                    });
-                    return;
-                  }
+                  await trackTransaction({
+                    txHash: tx.txHash,
+                    chainId: tx.chainId,
+                  });
                 } else if (abandoned) {
                   await onReindex(tx.txHash, tx.chainId);
                 }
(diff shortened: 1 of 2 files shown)


JunyupHong (commented):
### `state` vs `status` vs `released`

- **`state`**: end-to-end final state of one whole submitted tx (`STATE_COMPLETED_SUCCESS`, `STATE_ABANDONED`, etc.)
- **`status`**: status of a single hop within `transferSequence` (`pending`/`completed`/`failed`)
- **`released`**: only means "assets exited escrow/bridge custody and are accessible" — not completion. Per the type's own doc comment, assets can still be in transit even when `released === true`.

### Worth double-checking

```js
transactionStatuses[i + 1]?.transferAssetRelease?.released !== true
```

This treats "next tx not yet released" the same as "next tx stuck," but `released` can't tell those apart. Two possible false positives:

1. Next leg is still in flight (not released yet, but not failed either) → "Try again" could show too early.
2. Transfer types that never set `transferAssetRelease` (e.g. plain IBC) stay `released !== true` forever, even after they actually succeed → permanent misattribution to the prior leg.

**Suggestion**: check the next tx's `state` (or `getSimpleOverallStatus`) for an explicit failure instead of checking for absence of `released`:

```js
const nextStuck = next ? getSimpleOverallStatus(next.state) === "failed" : false;
```

mintae-clk:
Addressed the review.
**transferAssetRelease "Try again" logic**

1. Decide "stopped" by state, not released. released only means assets left escrow (true even mid-flight), so it misfired now a leg counts as stopped only when its tx state is actually failed.
2. Use slice(releasedTxIndex) instead of next.state. If the release happens on the last tx, there's no "next" to check — starting from the released tx covers the last one too.
3. Attribution by txIndex (provenance). Each event knows its tx, so the release lands on the right card even when the release chain differs from the next tx's origin (Injective).

JunyupHong (approved):

PR #1805, opened 2026-08-21
fix(widget): drop un-broadcast txs from explorer link (trailing comma)

mintae-clk:
## Problem
On multi-tx routes, the "View on explorer" link had a trailing comma and broke:

- `?tx_hash=0x912...6928,&chain_id=43114`

## Root cause
`createSkipExplorerLink` joined **every** `transactionDetails` entry, including txs
that haven't broadcast yet. A tx only gets a `txHash` once broadcast, so an
un-broadcast entry became `undefined` → `join(",")` leaked an empty value:

[tx1 (hash O), tx2 (hash X — stopped here)] → ["0x..", undefined] → "0x..,"

Since the route stopped at that tx, only the broadcasted txs should be in the link.

## Fix
`explorerLink.ts` — filter to txs with a `txHash` before join; return `""` when none
are broadcast.

This is the single shared entry point for every consumer (reverted / timeout /
history / failed), so the **root** fix covers them all. No consumer-side guards
needed — these pages are only reached after a tx is broadcast (`txHash` exists ⟹
broadcast; History's list filter requires a `txHash`), so `""` never reaches them.

## Result
`?tx_hash=0x912...6928&chain_id=43114` (no comma)

## Changeset
`@skip-go/widget` patch.


JunyupHong (approved):

JunyupHong:
There is no problem with the modified logic.
However, when there are multiple tx hashes and only one chainId as parameters to skip explorer, it seems it will not load properly.
This seems like a bug that has existed for a while, but if this can happen, then it might be worth considering encoding it with base64 as well.

